PR #29, opened 2026-04-06
docs: replace hardcoded devnet RPC URLs with placeholder

### LINKED ISSUE #21: docs: voting CLI examples hardcode devnet RPC
The `cast-vote`, `finalize-ballot`, and other voting CLI examples all reference `api.devnet.solana.com`. However, the `await-snapshot` flow is clearly production-ready (release mode, `/mnt/ledger` paths).

This creates a gap for mainnet operators who need to adjust RPC URLs. A simple note like "Replace with your mainnet RPC endpoint" or a `--rpc-url <YOUR_RPC>` placeholder would help.

Digitalenergyllc:
Closes #21

## Summary
All NCN CLI examples hardcoded `https://api.devnet.solana.com` as the RPC endpoint. The governance program is not deployed on devnet — these examples would fail for any operator trying to use them.

## Changes
- Replaced all devnet RPC URLs with `<RPC_URL>` placeholder in `docs/src/content/ncn/cli/index.mdx` (~22 instances)
- Same replacement in `ncn/README.md` (~13 instances)

Fixes #21

### DIFF
--- a/docs/src/content/ncn/cli/index.mdx
+++ b/docs/src/content/ncn/cli/index.mdx
@@ -184,7 +184,7 @@ Initializes the on-chain `ProgramConfig` singleton. Must be run once per deploym
 ncn-cli \
   --payer-path ~/.config/solana/id.json \
   --authority-path ~/.config/solana/id.json \
-  --rpc-url https://api.devnet.solana.com \
+  --rpc-url <RPC_URL> \
   init-program-config \
   --svmgov-program-id <PUBKEY>
 ```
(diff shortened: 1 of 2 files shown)


Digitalenergyllc:
Checked — the two remaining devnet references are in the globals reference table, not in any runnable example. Line 52 reproduces the --rpc-url help text, which lists mainnet, devnet, and a local validator as examples, and line 57 enumerates the valid --cluster values, one of which is devnet. Both match the #[arg] doc comments in ncn/cli/src/main.rs. Every command block now uses <RPC_URL>. Happy to drop the devnet example from line 52 if you'd prefer, but line 57 needs it to stay accurate.

tiago18c (approved):

PR #30, opened 2026-04-06
docs: add conceptual overview and quick start to root README

### LINKED ISSUE #18: docs: root README needs conceptual overview and getting-started guidance
The root README has a brief intro and component summaries but doesn't explain how `ncn/`, `svmgov/`, and `frontend/` connect or where an operator should start. The `docs/` directory is listed as a Nextra site but appears empty.

**Suggested improvements:**
- Add a high-level architecture paragraph explaining how the three components relate
- Add a "Getting Started" section that routes readers to the right subdirectory based on their role (operator, staker, developer)
- Either populate `docs/` or remove the reference to avoid confusion

Digitalenergyllc:
Closes #18

## Summary
The root README only listed project directories with no explanation of what the governance system does, how the pieces fit together, or how to get started.

## Changes
- Added one-paragraph overview explaining the two governance tracks (SVM + NCN)
- Added quick-start table linking to key docs by use case
- Added contributing section pointing to open issues
- Retained existing project structure section

## Note
This PR and #36 both touch the end of README.md — whichever merges second will need a trivial rebase, which I'll push same-day.

Fixes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,25 @@
 # Solana Validator Governance System
 
-A decentralized governance platform for Solana validators featuring merkle proof verification, stake-weighted voting, and real-time event monitoring.
+A decentralized governance platform enabling Solana validators and stakers to create, vote on, and finalize on-chain proposals using stake-weighted voting and merkle proof verification.
+
+## Overview
+
+The Solana Governance System provides two governance tracks:
+
+- **SVM Governance (`svmgov/`)** — Validator and staker proposals with stake-weighted voting, vote overrides for delegated stake, and on-chain finalization
+- **NCN Governance (`ncn/`)** — Node Consensus Network ballot voting with merkle proof verification, operator whitelisting, and a verifier service for automated snapshot processing
+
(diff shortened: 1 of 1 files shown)


Digitalenergyllc:
Fixed. The NCN track is not stake-weighted — ncn-snapshot's own README states that all whitelisted operators have equal voting weight, and cast_vote.rs increments the tally by one with no stake lookup. min_consensus_threshold_bps is a percentage of operators. Rewrote the overview to distinguish the two tracks: svmgov weights by stake with delegator override, while NCN establishes which snapshot is canonical and svmgov then uses it to weight proposal votes.

Unrelated, but noticed while checking: the ballot_tallies field comment in ncn/programs/ncn-snapshot/src/state/ballot_box.rs reads "Mapping of ballots votes to stake weight," which contradicts both the README and the handler. Happy to open a separate PR for that.

Digitalenergyllc:
Fixed in the latest commit — split into separate NCN operator CLI and svmgov CLI entries, since the two tracks have different tooling and a validator or staker following "Use the CLI" would land in operator snapshot commands.

tiago18c (approved):

PR #32, opened 2026-04-06
docs: clarify Rust 1.89.0 requirement and add dependency-patch warning

### LINKED ISSUE #20: docs: troubleshooting section recommends source-level patches to dependencies
The troubleshooting section recommends commenting out `panic`/`assert` lines in jito-solana in two instances:
1. Snapshot bank verification
2. Genesis creation time mismatch

While these don't impede merkle generation, recommending source-level patches to a dependency in production-facing documentation is a concern. An operator following these instructions could inadvertently apply these patches to a production validator binary.

**Suggested improvements:**
- Add a clear warning that these patches should NEVER be applied to production validator binaries
- Consider upstreaming fixes to the relevant repos
- If the patches are necessary for the governance tooling, document a separate build target or binary name to prevent confusion

### LINKED ISSUE #22: docs: Rust 1.89.0 requirement lacks explanation
Rust 1.89.0 is listed as a requirement but there's no explanation of why this specific version is needed or what breaks on other versions. Operators who already have a different Rust version installed need to understand whether this is a hard requirement (build fails) or a soft recommendation.

Digitalenergyllc:
Closes #22
Closes #20

## Summary
The NCN README pins Rust 1.89.0 as a prerequisite but doesn't explain why, and the troubleshooting section didn't warn that some workarounds modify dependency source code.

## Changes
- Added a note explaining the 1.89.0 requirement (dependency versions pinned by jito-tip-router and the Solana SDK; older toolchains fail to compile), with a `rustup override` tip to avoid changing the system default
- Added a warning to Troubleshooting that source-level dependency patches are fragile, lost on `cargo update`, and unsuitable for production (consolidated from #33)

Fixes #22

### DIFF
--- a/ncn/README.md
+++ b/ncn/README.md
@@ -85,6 +85,7 @@ rustup toolchain install 1.89.0 // install
 rustup default 1.89.0 // set as default
 rustc --version // verify version
 ```
+> **Why Rust 1.89.0?** The NCN program depends on nightly features used by the `jito-tip-router` and Solana SDK crates. Stable Rust will not compile these dependencies. Install with `rustup toolchain install 1.89.0` or set a directory override with `rustup override set 1.89.0`.
 
 2. (Optional - when using Anchor CLI) Install Solana CLI version 3.0 or higher. The bundled rustc in older Solana CLI versions may not be compatible with some dependencies.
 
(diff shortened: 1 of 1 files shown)


Digitalenergyllc:
Addressed in 038d3eb — reworded to present rustup override as an alternative rather than a competing instruction to rustup default.

tiago18c:
This is now stale given that we've already removed the jito dependency

Digitalenergyllc:
Agreed — with jito-tip-router removed from ncn/Cargo.toml upstream (nice cleanup by Prompt Logic), the dependency-patch warning and the epoch-rewards-hasher troubleshooting this PR added no longer apply. Closing. I'll close the linked issues as well.

PR #34, opened 2026-04-06
docs: add monitoring and alerting guidance for verifier service

Digitalenergyllc:
Closes #25

## Summary
The verifier service deployment docs explain setup but provide no guidance on monitoring, alerting, or crash recovery.

## Changes
- Documented the service's actual monitoring surfaces: /healthz, /version, /meta, and the token-gated /metrics endpoint (upload_total, proofs_not_found_total, free_storage_mb)
- Added an alerts table where every condition maps to an emitted signal, with notes that free_storage_mb is absolute MB and upload errors are cumulative counters (alert on deltas)
- Documented Docker logging and noted setup.sh's --restart unless-stopped as the default recovery path
- Scoped the systemd unit example to native-binary deployments, with the required EnvironmentFile for section-6 variables

Revised after review feedback to align all guidance with what src/metrics.rs and the deployment model actually support.

### DIFF
--- a/ncn/verifier-service/DEPLOYMENT.md
+++ b/ncn/verifier-service/DEPLOYMENT.md
@@ -176,3 +176,61 @@ Notes:
 
 - No changes are required for the cleanup cron; it runs on the host and continues to manage `/srv/verifier/data/governance.db`.
 - For zero-downtime, you can adapt the script to start a secondary container (different port) and flip traffic via a proxy/ALB once healthy.
+
+## Monitoring & Alerting
+
+### Health Checks
+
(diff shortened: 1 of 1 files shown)


Digitalenergyllc:
Fixed. The endpoint is /admin/stats, not /metrics — verified against current main (main.rs:121) and a live verifier, where /metrics 404s and /admin/stats 401s without the header. Also corrected the response shape: the storage fields are nested under storage, so it's storage.free_storage_mb, not top-level. Documented the 401 vs 503 distinction, enumerated the outcome and proof-kind values from metrics.rs, and noted that the counters are in-process and reset on restart.

MicaiahReid (approved):
LGTM!

PR #36, opened 2026-04-06
docs: add architecture diagram and glossary

Digitalenergyllc:
Closes #28

## Summary
New operators have no visual overview of how the governance system components connect. Understanding the system requires reading every individual README.

## Changes
- New `ARCHITECTURE.md` with:
  - Mermaid diagram showing component relationships (CLIs → programs, verifier → NCN, frontend → RPC)
  - Component overview tables for SVM and NCN governance tracks
  - End-to-end data flow walkthrough (epoch snapshot → merkle tree → on-chain upload → voting → finalization)
  - Glossary of 14 governance-specific terms
- Linked from root README

## Note
This PR and #30 both touch the end of README.md — whichever merges second will need a trivial rebase, which I'll push same-day.

Fixes #28

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,95 @@
+# Architecture & Glossary
+
+This document provides a visual overview of how the Solana Governance System components connect, and defines key terminology.
+
+## System Architecture
+
+```mermaid
+graph TB
+    subgraph "On-Chain Programs"
+        SVMGOV["svmgov Program
+(proposals, voting, finalization)"]
+        NCN["NCN Snapshot Program
+(ballots, merkle verification, whitelisting)"]
+    end
+
+    subgraph "Operator Tools"
+        SVMCLI["svmgov CLI"]
+        NCNCLI["NCN CLI"]
+        VS["Verifier Service"]
+        ROUTER["NCN Router
+(cron-based automation)"]
+    end
+
+    subgraph "User Interfaces"
+        FE["Frontend / Web UI"]
+    end
+
+    subgraph "Data Flow"
+        RPC["Solana RPC"]
+    end
+
+    SVMCLI -->|"create proposal / cast vote / finalize"| SVMGOV
+    NCNCLI -->|"cast vote / verify proof"| NCN
+    VS -->|"process snapshot → upload merkle proof"| NCN
+    ROUTER -->|"automated ballot management"| NCN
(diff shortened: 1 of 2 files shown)


Digitalenergyllc:
Fixed. The verifier service doesn't publish on-chain — it indexes operator-uploaded snapshots and serves proofs over HTTP (POST /upload, GET /proof/*). The merkle root reaches the chain when operators cast_vote into the BallotBox, and becomes canonical via finalize_ballot creating the ConsensusResult. Reversed the diagram edge and rewrote the data flow accordingly: each operator independently generates a snapshot and votes its own root, which is the point of the design — there's no single uploader. Also corrected "votes weighted by stake" in the aggregation step and the Quorum glossary entry; NCN operators have equal voting weight per ncn-snapshot's README, and the threshold is min_consensus_threshold_bps over operators. Verified against a mainnet run at slot 437036500.

Digitalenergyllc:
Added — the frontend now shows a write path to svmgov for vote, support, and override transactions, alongside the existing read path via RPC and proof retrieval from the verifier service.

Digitalenergyllc:
Fixed — split the edge. The CLI generates the snapshot and signs the hash via log-meta-merkle-hash; the operator then makes a separate signed multipart POST to the verifier's /upload. There's no upload subcommand in the CLI.

MicaiahReid (changes requested):
@Digitalenergyllc, could you resolve conflicts?

Also, this has the NCN Router as an Operator Tool, running "automated ballet management". Alongside that, the verifier is "serves proofs to voters and UIs".

While the verifier does server proofs, it's the router service, run by foundation, that checks the validity of and whitelists the operator's running verifiers and servers the proofs by redirecting to one of those verifiers.

I think it's worth showing that - the NCN router running outside of the "Operator Tools", and sitting between the Verifier and Frontend. Thoughts?

Digitalenergyllc:
@MicaiahReid Agreed and done — moved the NCN Router out of Operator Tools into its own foundation-run grouping between the frontend and the operators' verifiers: proof requests hit the router, which validates/whitelists the operator verifiers and redirects to a healthy one. Updated the component table, data-flow step 5, and added a glossary entry to match. Also rebased onto current main with the README conflict resolved. Done in efc8d87.

MicaiahReid (approved):
Beautiful, thanks @Digitalenergyllc!

PR #55, opened 2026-05-28
docs: snapshot-slot recovery + generate-meta-merkle path clarification

Digitalenergyllc:
Documents the snapshot-slot command and how to recover when await-snapshot can't find an incremental snapshot covering the target slot — it replays to build the exact-slot snapshot that generate-meta-merkle requires.
Complements #52, which clarifies the await message; this covers the recovery procedure the message points toward. Field-tested on a live mainnet round: a single-pass ~84k-slot replay produced a MetaMerkle root and snapshot hash matching cohort consensus exactly. Also notes the retention config that prevents the gap.
Also clarifies that generate-meta-merkle reads the full snapshot from --backup-snapshots-dir, not --full-snapshots-path — prevents the common "Missing snapshot at slot X" confusion.

### DIFF
--- a/docs/src/content/ncn/cli/index.mdx
+++ b/docs/src/content/ncn/cli/index.mdx
@@ -65,12 +65,13 @@ These options are available on all commands. Each can also be set via the matchi
 
 ### `snapshot-slot`
 
-Generates a Solana ledger snapshot for a specific target slot. Replays the validator bank up to `--slot` and writes a full snapshot into `--backup-snapshots-dir`. Requires `--ledger-path` and `--full-snapshots-path` (or defaults derived from `--ledger-path`).
+Generates a Solana ledger snapshot for a specific target slot. Replays the validator bank up to `--slot` and writes a full snapshot into `--backup-snapshots-dir`. Requires `--ledger-path` and `--full-snapshots-path` (or defaults derived from `--ledger-path`). Unlike `generate-meta-merkle` — which requires an exact-slot snapshot to already exist — `snapshot-slot` creates one, making it the recovery path when `await-snapshot` cannot find snapshot coverage for the target slot (see below).
 
 ```bash
 ncn-cli \
   --ledger-path /mnt/ledger \
   --full-snapshots-path /mnt/ledger/full-snapshots \
+  --account-paths /mnt/accounts \
   --backup-snapshots-dir /mnt/ledger/backup-snapshots \
   snapshot-slot --slot <SLOT>
 ```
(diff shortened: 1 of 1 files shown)


Digitalenergyllc:
Heads up before you dive in — I'm pushing an update to this PR today folding in the operational findings from this week's mainnet runs (the genesis-copy requirement, the snapshot retention window, the real disk budget, and a symptom→cause→fix failure table), per my note in the TG. Worth holding review until that lands — I'll comment here when it's up.

Digitalenergyllc:
Update is in as c239c1f — the genesis-copy requirement, the real disk budget (~350GB observed), and a symptom→cause→fix quick-reference covering every failure mode from this week's mainnet runs. The retention-window guidance was already in the earlier commit. Ready for review.

docs/src/content/ncn/cli/index.mdx
```diff
@@ -92,6 +85,39 @@ ncn-cli \
...
+>
+> Retain enough incrementals that any target slot has one terminating at or before it. A working baseline: full retention `2` / full interval `100000` / incremental retention `30` / incremental interval `2500` — roughly 75,000 slots of incremental coverage, which keeps `await-snapshot` on its happy path.
+
+> **`await-snapshot`: stage genesis into the backup ledger dir first**
+>
+> The internal `agave-ledger-tool ... copy` step populates rocksdb into `--backup-ledger-dir` but does **not** copy the genesis files. The subsequent snapshot-creation step then fails with a misleading `Missing snapshot at slot <X>`, preceded by a `Failed to load genesis config` line. Before running `await-snapshot`, copy `genesis.tar.bz2` and `genesis.bin` from the live ledger directory into `--backup-ledger-dir`.
+
+> **Disk budget**
+>
+> Budget roughly **350GB free** for a mainnet run: a ~118GB full-snapshot copy, ~90GB of copied ledger range (rocksdb), the accounts unpack during bank load, and the newly created target-slot snapshot. 150GB is not enough — runs fail mid-flight.
```
MicaiahReid: Is there any reason this whole section is quoted with `>`? Looks weird in the final rendering

MicaiahReid (changes requested):

Digitalenergyllc:
Fair point — a stack of blockquotes renders poorly on the docs site. Converted the whole troubleshooting section to plain #### subsections, and did the same to the matching "Full-snapshot lookup" note under generate-meta-merkle for consistency. Left the repo's pre-existing callouts as-is. Done in 646460f.

MicaiahReid (approved):
Nice, thanks @Digitalenergyllc!

PR #72, opened 2026-06-15
fix(svmgov): prevent proposal backdating in flush_merkle_root and support_proposal

MicaiahReid:
## Summary

Fixes a proposal **backdating** vulnerability in the svmgov program.

`flush_merkle_root` recomputed a supported proposal's snapshot lineage and then
skipped the `ncn_snapshot::init_ballot_box` CPI whenever the supplied
`ballot_box` account was non-empty (`if self.ballot_box.data_is_empty()`).
Because `init_ballot_box` held the **only** `snapshot_slot > clock.slot` guard,
a proposal author could:

1. Pass an arbitrary non-empty account (e.g. another `Proposal` PDA) as
   `ballot_box`, skipping the CPI and its validation, and
2. Have the proposal's `snapshot_slot` / `consensus_result` recomputed onto an
   **already-past** slot when config (`snapshot_epoch_extension` /
   `snapshot_slot_offset`) lets the recomputed slot fall in the past.

The proposal then rebinds onto an already-finalized `ConsensusResult`, and
`cast_vote` later accepts proofs against that already-known meta-merkle root.

The sibling instruction `support_proposal` establishes the same lineage through
the identical unchecked-`ballot_box` / skip-CPI sink, so it is hardened the same
way to satisfy the invariant rather than only the reported entry point.

## Fix

For **both** lineage-setting instructions (`flush_merkle_root` and
`support_proposal`):

- **Enforce `snapshot_slot > clock.slot` before mutating any proposal state**,
  via a new `compute_future_snapshot_slot` helper. This mirrors the guard inside
  `ncn_snapshot::init_ballot_box`, so the future-slot invariant holds even when
  the CPI is skipped.
- **Bind `ballot_box` to the exact PDA** derived from the recomputed
  `snapshot_slot` (`require_keys_eq!` against
  `["BallotBox", snapshot_slot.to_le_bytes()]`), rejecting any arbitrary account
  used to skip the CPI.
- **Reorder writes** to `snapshot_slot`, `start_epoch`, `end_epoch`, and
  `consensus_result` so they only happen **after** validation succeeds.

Adds the `InvalidBallotBox` error (appended to the enum to preserve existing
error-code numbers) and updates the committed IDLs accordingly.

## Compatibility

No legitimate flow is affected. Any honest call already targets a future
snapshot slot — the `init_ballot_box` CPI would otherwise reject a past slot at
creation time — and honest callers always pass the correct `BallotBox` PDA
(required for the `init` CPI to succeed). The new checks only reject the exploit
path (skipping the CPI with a stale slot and/or a mismatched `ballot_box`), so
no previously valid request becomes invalid.

## Testing

- `cargo test -p svmgov_program --lib` — added unit tests for
  `compute_future_snapshot_slot` covering the accepted future-slot case, the
  positive-offset case, equal-to-current rejection, past-slot rejection,
  negative-offset underflow, and the exploit's
  backdating-via-negative-offset rejection. All 8 lib tests pass.
- `cargo build -p svmgov_program` — clean, no new warnings.

### DIFF
--- a/svmgov/program/programs/svmgov_program/src/utils.rs
+++ b/svmgov/program/programs/svmgov_program/src/utils.rs
@@ -157,3 +157,106 @@ pub fn get_epoch_slot_range(epoch: u64) -> (u64, u64) {
 
     (start_slot, end_slot)
 }
+
+/// Computes the snapshot slot for a proposal's voting lineage from the snapshot
+/// `target_epoch` and the configured `snapshot_slot_offset`, enforcing that the
+/// resulting slot is strictly in the future relative to `current_slot`.
+///
+/// The `snapshot_slot > current_slot` invariant mirrors the guard inside
+/// `ncn_snapshot::init_ballot_box`. Both `support_proposal` and
+/// `flush_merkle_root` skip that CPI whenever the supplied `ballot_box` account
+/// already exists, so without re-checking here a caller could bind a proposal to
+/// an already-finalized `ConsensusResult` for a past slot (proposal backdating).
+///
+/// Returns the validated `snapshot_slot`, or an error if the offset underflows
+/// below zero or the resulting slot is not in the future.
+pub fn compute_future_snapshot_slot(
+    target_epoch: u64,
+    snapshot_slot_offset: i64,
+    current_slot: u64,
+) -> core::result::Result<u64, crate::error::GovernanceError> {
+    let (start_slot, _) = get_epoch_slot_range(target_epoch);
+    let offset_result = (start_slot as i64)
+        .checked_add(snapshot_slot_offset)
+        .ok_or(crate::error::GovernanceError::ArithmeticOverflow)?;
+    if offset_result < 0 {
+        return Err(crate::error::GovernanceError::ArithmeticOverflow);
+    }
+    let snapshot_slot = offset_result as u64;
+    if snapshot_slot <= current_slot {
+        return Err(crate::error::GovernanceError::InvalidSnapshotSlot);
+    }
+    Ok(snapshot_slot)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::error::GovernanceError;
+
+    #[test]
+    fn epoch_slot_range_is_correct() {
+        assert_eq!(get_epoch_slot_range(0), (0, 431_999));
+        assert_eq!(get_epoch_slot_range(1), (432_000, 863_999));
+    }
+
+    #[test]
+    fn future_snapshot_slot_accepts_future_slot() {
+        // Epoch 2 starts at slot 864_000; current slot is well before that.
+        assert_eq!(
+            compute_future_snapshot_slot(2, 0, 500_000).unwrap(),
+            864_000
+        );
+    }
+
+    #[test]
+    fn future_snapshot_slot_applies_positive_offset() {
+        assert_eq!(
+            compute_future_snapshot_slot(2, 100, 500_000).unwrap(),
+            864_100
+        );
+    }
+
+    #[test]
+    fn future_snapshot_slot_rejects_past_slot() {
+        // Recomputed snapshot slot (864_000) is behind the current slot.
+        assert!(matches!(
+            compute_future_snapshot_slot(2, 0, 900_000),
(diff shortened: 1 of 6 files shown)


MicaiahReid:
@cantina-integration fix review
Please verify this fix against the original finding and the Apex Fix Requirements.

svmgov/program/programs/svmgov_program/src/utils.rs
```diff
@@ -157,3 +157,106 @@ pub fn get_epoch_slot_range(epoch: u64) -> (u64, u64) {
...
+/// below zero or the resulting slot is not in the future.
+pub fn compute_future_snapshot_slot(
+    target_epoch: u64,
+    snapshot_slot_offset: i64,
+    current_slot: u64,
+) -> core::result::Result<u64, crate::error::GovernanceError> {
+    let (start_slot, _) = get_epoch_slot_range(target_epoch);
+    let offset_result = (start_slot as i64)
+        .checked_add(snapshot_slot_offset)
+        .ok_or(crate::error::GovernanceError::ArithmeticOverflow)?;
```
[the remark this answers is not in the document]
MicaiahReid: Not a concern

svmgov/program/programs/svmgov_program/src/utils.rs
```diff
@@ -157,3 +157,106 @@ pub fn get_epoch_slot_range(epoch: u64) -> (u64, u64) {
...
+    if offset_result < 0 {
+        return Err(crate::error::GovernanceError::ArithmeticOverflow);
+    }
+    let snapshot_slot = offset_result as u64;
+    if snapshot_slot <= current_slot {
+        return Err(crate::error::GovernanceError::InvalidSnapshotSlot);
+    }
```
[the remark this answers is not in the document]
MicaiahReid: Fixed in a9a866da75b5558be12e466b2e07c86ec390939a

svmgov/program/programs/svmgov_program/src/utils.rs
```diff
@@ -157,3 +157,106 @@ pub fn get_epoch_slot_range(epoch: u64) -> (u64, u64) {
...
+/// Returns the validated `snapshot_slot`, or an error if the offset underflows
+/// below zero or the resulting slot is not in the future.
+pub fn compute_future_snapshot_slot(
+    target_epoch: u64,
+    snapshot_slot_offset: i64,
+    current_slot: u64,
+) -> core::result::Result<u64, crate::error::GovernanceError> {
+    let (start_slot, _) = get_epoch_slot_range(target_epoch);
+    let offset_result = (start_slot as i64)
+        .checked_add(snapshot_slot_offset)
```
dhruvsol, line 180: will this work for negative slots offset ? the idea with keeping offset as i64 was to allow negative offset as the math for slots doesn't match on testnet because of restarts or some other factors, the same case can happen on mainnet in future. thoughts ?

MicaiahReid: I'm not fully following, can you elaborate? Also note, this is keeping the original arithmetic done by `flush_merkle_root` in tact, it's just functionizing it so it can reuse the added check against `current_slot`.

The original was:
```rust
        let (start_slot, _) = get_epoch_slot_range(target_epoch);
        let offset_result = (start_slot as i64)
            .checked_add(self.global_config.snapshot_slot_offset)
            .ok_or(GovernanceError::ArithmeticOverflow)?;
        require!(offset_result >= 0, GovernanceError::ArithmeticOverflow);
```

dhruvsol: my bad i missed "as i64", this will work fine

about the negative offset so, testnet epoch slot math wont result in correct slot number because of cluster restart there is a small offset so when we were testing on testnet we were using a negative offset which was allowing us to ping snapshot slot to epoch + 1000 slots, if you do the current math with same numbers on testnet it will result in a wrong snapshot slot.

MicaiahReid: Nice, this negative offset functionality should still be possible from what I can tell!

dhruvsol (approved):

PR #74, opened 2026-06-15
fix(svmgov): bind flush_merkle_root to the support-derived schedule

MicaiahReid:
## Summary

Fixes an author-controlled scheduling vulnerability in `flush_merkle_root`.

`flush_merkle_root` recomputed a supported proposal's `snapshot_slot`,
`start_epoch`, and `end_epoch` from the **current** epoch
(`clock.epoch + snapshot_epoch_extension`) and **omitted `discussion_epochs`**.
Because it re-anchored on the current epoch every call and never revalidated
support or proposer eligibility, a proposal author could:

1. Call `flush` immediately after support to **drop the configured
   `discussion_epochs` window** (shortening time-to-vote), and
2. Call `flush` repeatedly in later epochs to **push `start_epoch` forward
   indefinitely** before voting begins — while `cluster_support_lamports`
   stayed frozen (stale sponsorship).

## Fix

Anchor the flush schedule on the **immutable, support-derived epoch** instead of
the current epoch:

```
target_epoch = creation_epoch + max_support_epochs   // the support epoch
             + discussion_epochs                      // discussion window retained
             + snapshot_epoch_extension
```

`creation_epoch` is set once at creation and never changes, so the recomputed
`snapshot_slot` / `start_epoch` / `end_epoch` are now deterministic. This is
exactly the schedule `support_proposal` already computed, so `flush` reconstructs
the original schedule rather than letting the author re-pick it. It can no longer
shorten the discussion period or postpone `start_epoch`.

The root cause was that `flush`'s schedule formula had **drifted** from
`support_proposal`'s. To prevent that recurring, both instructions now derive the
schedule through one shared `proposal_target_epoch()` helper:

- `support_proposal` passes `clock.epoch` — **behavior-preserving**, because the
  instruction already requires `clock.epoch == creation_epoch + max_support_epochs`
  at that point, so the computed value is unchanged.
- `flush_merkle_root` passes `creation_epoch + max_support_epochs`.

## Compatibility / behavior

- `support_proposal`: no behavioral change (same computed schedule, refactored to
  the shared helper).
- `flush_merkle_root`: now reconstructs the original support-derived schedule.
  Re-flushing before the snapshot slot arrives re-derives the same
  `consensus_result` and (re)creates the ballot box for the original schedule —
  the legitimate "retry the snapshot" use. It no longer reschedules the proposal;
  rescheduling would require re-creating/re-supporting the proposal, which is the
  intended behavior per the finding.

## Testing

- `cargo test -p svmgov_program --lib` — added unit tests asserting `flush` and
  `support` derive the **same** target epoch from their respective anchors, and
  that the `discussion_epochs` window is retained in the schedule. All 10 lib
  tests pass.
- `cargo build -p svmgov_program` — clean, no new warnings.

---

### ⚠️ Stacking / review notes

- **Stacked on #72** (`fix/flush-merkle-root-backdating`). This PR's base is that
  branch, so the diff shows only this finding's changes. When #72 merges, GitHub
  will retarget this PR's base to `main`.
- **Apex marker pending:** the finding hand-off for this issue did not include an
  `apex-fix-review` marker token or the `@cantina-integration` start-comment text,
  so neither has been added yet. Once provided, I'll add the marker to this body
  and post the start comment.


svmgov/program/programs/svmgov_program/src/utils.rs
```diff
@@ -158,6 +158,25 @@ pub fn get_epoch_slot_range(epoch: u64) -> (u64, u64) {
...
+/// anchored on the *current* epoch (as it previously did) it would drop the
+/// `discussion_epochs` window and let the author push `start_epoch` forward
+/// indefinitely before voting begins, without revalidating support.
+pub fn proposal_target_epoch(
+    support_epoch: u64,
+    discussion_epochs: u64,
+    snapshot_epoch_extension: u64,
+) -> u64 {
+    support_epoch + discussion_epochs + snapshot_epoch_extension
+}
```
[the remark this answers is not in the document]
MicaiahReid: Fixed: c1749ad743f39730090530f466ca955e95197770

svmgov/program/programs/svmgov_program/src/instructions/flush_merkle_root.rs
```diff
@@ -68,9 +68,19 @@ impl<'info> FlushMerkleRoot<'info> {
...
-        // Using the same logic as in support_proposal
-        let target_epoch = clock.epoch + self.global_config.snapshot_epoch_extension;
+        // SECURITY: reconstruct the *original*, support-derived schedule rather than
+        // re-anchoring on the current epoch. `support_proposal` set the schedule from
+        // the support epoch (creation_epoch + max_support_epochs) plus discussion_epochs
+        // and snapshot_epoch_extension. Anchoring on the immutable creation_epoch makes
+        // flush deterministic: re-flushing never drops the discussion window and never
+        // pushes start_epoch forward, so a flush cannot shorten or repeatedly postpone
+        // the schedule without revalidating support.
+        let support_epoch = self.proposal.creation_epoch + self.global_config.max_support_epochs;
```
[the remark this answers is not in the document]
MicaiahReid: Fixed c1749ad743f39730090530f466ca955e95197770

svmgov/program/programs/svmgov_program/src/instructions/flush_merkle_root.rs
```diff
@@ -68,9 +68,23 @@ impl<'info> FlushMerkleRoot<'info> {
...
+        // SECURITY: reconstruct the *original*, support-derived schedule rather than
+        // re-anchoring on the current epoch. `support_proposal` set the schedule from
+        // the support epoch (creation_epoch + max_support_epochs) plus discussion_epochs
+        // and snapshot_epoch_extension. Anchoring on the immutable creation_epoch makes
+        // flush deterministic: re-flushing never drops the discussion window and never
+        // pushes start_epoch forward, so a flush cannot shorten or repeatedly postpone
+        // the schedule without revalidating support.
+        let support_epoch = self
+            .proposal
+            .creation_epoch
```
dhruvsol: this be the current epoch?

Also, the target epoch should not include discussion epochs, since it was decided that the discussion period was over. There was an issue with the snapshots, so the only reason to use flush is if NCN consensus failed and we do not have a valid snapshot for voting. In that case, we can redo it in the next epoch without any issues.

MicaiahReid: So the original problem described is that this ix (on line 119) sets `self.proposal.start_epoch = target_epoch + 1;`. Since `target_epoch` was a function of the current epoch, this `flush_merkle_root` could be called again and again to keep pushing `start_epoch` back. 

But you're saying part of the intent of this instruction _is_ to push back the `start_epoch`? If so, do we need additional restrictions on who can call this? 

>  the only reason to use flush is if NCN consensus failed

From what I understand, while that may be true, could this endpoint still be used to postpone `start_epoch` and keep voting from ever happening.

dhruvsol: yes, i think the start epoch push back is a problem but for a proposal to reach here at snapshot phase is very hard so we dont need to worry about people trying to spam the contract, support of 15% of cluster saves us here. also i doubt any major figure will try to push the proposals as the owner themselves can only do it

MicaiahReid: Per our conversation, to keep the ability to push back the start_epoch when NCN consensus isn't reached, but to prevent a proposer abusing this ability, we decided to make the flush merkle root endpoint require admin approval. 34ae47abdbc119c13bc56303930882411ae6860a

dhruvsol (approved):

PR #75, opened 2026-06-15
fix(frontend): set MetaMerkleProof close_timestamp to vote expiry

MicaiahReid:
The shipped frontend initialized MetaMerkleProof with close_timestamp = 1, which makes the proof permissionlessly closable immediately (the on-chain close gate only requires now >= close_timestamp). Because the override flows sent proof creation and the vote in two separate transactions, an attacker could close the proof between them and force the vote to fail, suppressing delegator override voting through the web app. The same hard-coded value affected the cast/modify vote flows.

Compute close_timestamp as the proposal's vote-expiry instant, mirroring the svmgov CLI's compute_vote_expiry_timestamp, so the proof cannot be closed while voting is open. Also bundle proof creation and the override vote into a single atomic transaction (the cast/modify vote flows already did this), as defense-in-depth so no inter-transaction window remains.

Add unit tests for the new close-timestamp computation.

### DIFF
--- a/frontend/src/chain/instructions/helpers.ts
+++ b/frontend/src/chain/instructions/helpers.ts
@@ -216,6 +216,81 @@ export function getMetaMerkleProofPda(
   return metaMerkleProofPda;
 }
 
+// Milliseconds per slot (matches solana_sdk::clock::DEFAULT_MS_PER_SLOT and the svmgov CLI's
+// compute_vote_expiry_timestamp).
+const MS_PER_SLOT = 400;
+
+/**
+ * Estimate the Unix timestamp (in seconds) at which voting expires for a proposal — i.e. the
+ * start of `endEpoch`, after which voting is no longer valid. This is the value a newly created
+ * `MetaMerkleProof` should use for its `close_timestamp`: until this time the proof cannot be
+ * closed permissionlessly, so it survives for the whole voting window and an attacker cannot
+ * delete the shared proof state between the init and vote steps. Mirrors the svmgov CLI's
+ * `compute_vote_expiry_timestamp`.
+ *
+ * There is no on-chain expiry timestamp (the proposal only stores `end_epoch`), so this estimates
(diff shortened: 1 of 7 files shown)


frontend/src/chain/instructions/helpers.ts
```diff
@@ -216,6 +216,81 @@ export function getMetaMerkleProofPda(
...
+    }
+    if (candidate === 0) break;
+    candidate = Math.max(0, candidate - 1);
+  }
+
+  throw new Error(
+    `Failed to fetch a recent block time (tried ${MAX_ATTEMPTS} slots ending at ${candidate}): ${String(
+      lastErr
+    )}`
+  );
```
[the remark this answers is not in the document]
MicaiahReid: Fixed: 20604a45b0370a87d86e8035bd0ae58d1c2ee485

frontend/src/chain/instructions/helpers.ts
```diff
@@ -216,6 +216,85 @@ export function getMetaMerkleProofPda(
...
+  // used as the reference, so the slot delta (and thus the estimate) stays consistent.
+  const [refSlot, refTime] = await blockTimeAtOrBefore(
+    connection,
+    info.absoluteSlot
+  );
+  const slotDelta = targetSlot - refSlot;
+
+  // Math.trunc (toward zero) matches Rust's i64 integer division in the CLI helper, so an
+  // already-expired proposal yields the same past timestamp in both code paths.
+  return refTime + Math.trunc((slotDelta * MS_PER_SLOT) / 1000);
```
dhruvsol: i think we should add some buffer ? in case the block we use by the validator is shorter or longer we might end up with wrong timining ? because then it will be like closing and reopening the merkle proof, which is not part of the flow

MicaiahReid: Fixed (also fixed for rust version of this fn): 84d09fe868ffb0f8bfb318374c9960298a0da187

dhruvsol (approved):

PR #84, opened 2026-06-17
fix(ncn-router): dedupe whitelist by canonical origin (slash-variant follow-up to #82)

MicaiahReid:
## Summary

Follow-up to the merged fix for the duplicate-routing-ticket finding (#82). A rescan found the remediation was incomplete: both the cron-side whitelist dedup and the router-side `select_routable_verifiers` keyed on the **raw `domain` string**, while `ncn-router` redirects to a verifier base only **after ensuring a trailing slash**. So `http://evil` and `http://evil/` survived as two separate `ok` rows (two routing tickets) yet collapsed to the same upstream after selection — restoring extra routing weight for a single origin.

## Fix

Introduce a shared `canonical_domain(domain)` helper that mirrors the router's redirect identity (ensure a single trailing slash, no other normalization) and key the de-duplication on it at both layers:

- **cron** (`cron_job.rs`): `dedupe_verifiers_by_domain` now collapses entries by canonical domain before writing the snapshot.
- **router** (`router.rs`): `select_routable_verifiers` de-duplicates the `ok` set by canonical domain before sampling, and the redirect `target` is built from the same canonical form so the routing identity is consistent end to end.

This collapses **exactly** the variants the router would collapse (trailing-slash) and no more, so genuinely-distinct origins (different host/port/path) are unaffected — no over-merging, no rejected-input compatibility impact.

## Why match the router exactly (no trimming, etc.)

The router does not trim whitespace; a whitespace variant like `http://evil ` produces a *different* (broken) upstream target, so it cannot grant extra weight to the same working upstream. Canonicalizing on trailing-slash only — the precise set the router collapses — closes the exploitable gap without merging entries that resolve to distinct upstreams.

## Testing

- `cargo test --bins`: 18 tests pass (5 cron + 13 router), including new `dedupe_collapses_trailing_slash_domain_variants` (cron) and `routable_verifiers_dedupe_trailing_slash_variants` (router).
- Existing network-binding, ok-preference, and percent-encoding tests remain green.

### DIFF
--- a/ncn-router/src/cron_job.rs
+++ b/ncn-router/src/cron_job.rs
@@ -371,24 +371,41 @@ fn compare_with_chain(
     Ok(())
 }
 
-/// Keep one canonical whitelist record per verifier origin (`domain`).
+/// Canonical routing identity for a verifier origin. `ncn-router` redirects to
+/// `domain` after ensuring a trailing slash, so `http://x` and `http://x/`
+/// resolve to the same upstream. De-duplicating on this canonical form (rather
+/// than the raw string) prevents slash variants from surviving as separate
+/// whitelist rows that the router would later collapse to one upstream — i.e.
+/// extra routing tickets for a single origin. Matches `canonical_domain` in
+/// `ncn-router`'s `router.rs`.
+fn canonical_domain(domain: &str) -> String {
+    if domain.ends_with('/') {
+        domain.to_string()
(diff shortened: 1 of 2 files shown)


MicaiahReid:
@cantina-integration fix review
Please verify this fix against the original finding and the Apex Fix Requirements.

PR #88, opened 2026-06-18
fix(ncn-router): enforce HTTPS for verifier domains

MicaiahReid:


MicaiahReid:
Don't know if we want to merge this one yet, but at some point we should probably require all verifiers be https

PR #93, opened 2026-06-26
fix: security.txt macro added to ncn,svmgov

tigarcia:


dhruvsol:
Hey @tigarcia we can close this ? i will setup program metadata so we dont break the build

PR #104, opened 2026-07-10
Multi epoch support validator limit

tigarcia:
Support was previously accepted in exactly one epoch (creation_epoch + max_support_epochs, enforced by strict equality), contradicting the documented meaning of max_support_epochs and the CLI's phase display. Support is now valid for the entire inclusive window [creation_epoch, creation_epoch + max_support_epochs].

Because supporters can now arrive in different epochs, stake readings from earlier epochs are never reused: each support call rebuilds the entire tally by re-measuring every prior supporter at the current epoch via the sol_get_epoch_stake syscall, so the numerator and denominator (get_epoch_total_stake) always come from the same epoch.

Changes:

    Proposal gains a supporters: Vec (last field, Borsh append-only); the account grows 32 bytes per supporter via realloc, rent paid by that supporter. NOTE: layout break — proposals created before this upgrade no longer deserialize.
    support_proposal: window gate (repurposes the previously unused SupportPeriodExpired error), full re-tally, realloc constraint.
    New permissionless retally_support instruction: re-measures support at the current epoch and activates voting if the threshold is met, covering stake drift when no new supporter arrives to trigger a re-tally. Emits ProposalRetallied.
    Voting activation (schedule + ballot-box PDA binding + CPI) extracted into a shared activate_voting used by both instructions.
    The voting schedule now anchors on whichever epoch crosses the threshold, so early support means an earlier vote.
    Pure helpers (check_support_window, tally_supporter_stakes, min_stake_threshold) in utils.rs with 14 unit tests.


cavemanloverboy (approved):
lgtm other than the known heap limitation w/ supporters. given the 2000 key limit, may be better to just pay 0.44 SOL up front then the original payer can reclaim after proposal ends. simplifes realloc logic.

tiago18c:
@cavemanloverboy this fixes the heap limitation https://github.com/solana-foundation/solana-governance/pull/106 , can you take a look?

tiago18c (approved):

PR #111, opened 2026-07-31
chore: version bumps + idl syncs + frontend test fix

tiago18c:


docs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "svmgov-docs",
-  "version": "0.3.0-40000",
+  "version": "0.4.0-40000",
```
tigarcia: I think the 40000 version corresponds to the agave version we depend on for the NCN

tigarcia (approved):

PR #116, opened 2026-08-03
feat(svmgov cli): add --with-support flag on create proposal to immediately support the proposal on creation

tiago18c:


MicaiahReid (dismissed):
I would support having a function in `svmgov/cli/src/instructions/support_proposal` that returns the instruction, which both `create_proposal` and `support_proposal` use, to save on duplicated code. 

But not worth holding up over, it's up to you :)

MicaiahReid:
Can approve once rebased @tiago18c!

MicaiahReid (approved):
LGTM!

PR #119, opened 2026-08-03
chore(svmgov): Request a Compute Budget For `support_proposal` and `retally_support`

0xIchigo:
## TLDR

`support_proposal` and `retally_support` re-tally the whole supporter list on every call, so their cost is linear in `num_supporters`, but neither client requested a compute budget at all, leaving them on the 200k default. This sizes the request from the actual supporter count instead of a flat worst case

## Background

Both handlers rebuild the entire supporter tally on every call: one `sol_get_epoch_stake` syscall per existing supporter, up to the `MAX_SUPPORTERS_LIMIT` of 2000. A transaction carrying only the governance instruction is allotted 200k CU, which does not cover that range. The program's own LiteSVM harness has always requested 1.4M for these two instructions; neither client requested anything

Measured: the 200k default runs out at 1,316 supporters. Mainnet has ~700 validators, so this is headroom rather than a live failure, but the clients could not service the supporter range the program advertises

## The model

Per review feedback, the request now scales with the supporter:

```
(22_500 + 132 × num_supporters + 28_000) × 1.15,  capped at 1_400_000
```

| supporters | requested | measured | headroom |
|---|---|---|---|
| 0 (activating, 64 operators) | 58,075 | 49,473 | 17% |
| 100 | 73,255 | — | — |
| 500 | 133,975 | 86,440 | 55% |
| 2,000 (cap) | 361,675 | 287,953 | 26% |

Both clients read `num_supporters` from the proposal account before building the transaction. The 15% margin chiefly covers supporters landing between that read and execution, at ~132 CU each, roughly 100+ supporters of slack even at the cap, so the race is not realistic

## The activation term

The `28_000` is the interesting part. The original measurement pre-seeded the ballot box, so `init_ballot_box` never ran, but that CPI fires on the threshold-crossing call, which is precisely the call the budget exists to protect. `ncn_flow` exercises the real path:

- support with no prior supporters, no activation: **22,434 CU**
- support that activates and creates the ballot box: **49,473 CU**

So activation adds ~26.8k. It is measured at the 64-operator whitelist maximum, because `init_ballot_box` clones the whitelist into the ballot box and a longer list costs more to serialize; an earlier draft used 25,000 from measurements at ≤8 operators, which quietly eroded the stated 15% margin to ~10%

The term is always included: a caller cannot know in advance whether its own call will be the one that crosses

## Changes

- `svmgov/cli/src/constants.rs` — the model plus `support_compute_unit_limit()`
- `svmgov/cli/src/instructions/{support_proposal,retally_support}.rs` — fetch the proposal, request the modelled limit
- `frontend/src/chain/instructions/types.ts` — `supportComputeUnitLimit()` mirror
- `frontend/src/chain/instructions/supportProposal.ts` — same, via a proposal fetch

`cast_vote` / `cast_vote_override` are deliberately unchanged: they verify a single merkle leaf rather than walking the list, and the harness correspondingly does not raise their limit. The frontend has no retally flow

## Tests

The model lives in three places (Rust CLI, TypeScript, and a test mirror in `tests/common/mod.rs` shared by both program suites). Cross-pinned values keep them honest: both client suites assert `supportComputeUnitLimit(0) == 58_075` and `(2_000) == 361_675`, so drift on either side fails CI

- **`tests/support_compute_budget.rs`** — proves the 200k default really is insufficient at the cap (so the request is load-bearing), then fills the list to 2000 and asserts the measured peak fits under what the clients request. It recomputes the clients' formula rather than comparing against a constant the same suite justifies
- **`tests/ncn_flow.rs`** — new `activation_cpi_fits_within_the_modelled_compute_limit` measures the activating support through the real CPI at the 64-operator maximum and asserts the model covers it. This is what keeps `28_000` from being an unexplained magic number
- **`svmgov/cli`** — unit tests for the formula: covers measured cost, scales with the count, clamps at 1.4M, matches the TypeScript mirror

Full program suite green (12 targets), CLI 11 unit + 3 integration, frontend 80 tests, `tsc` and eslint clean, touched Rust files rustfmt-clean

### DIFF
--- a/frontend/src/chain/instructions/supportProposal.ts
+++ b/frontend/src/chain/instructions/supportProposal.ts
@@ -11,6 +11,7 @@ import {
   SupportProposalParams,
   TransactionResult,
   SNAPSHOT_PROGRAM_ID,
+  MAX_SUPPORTERS,
   supportComputeUnitLimit,
   ChainVoteAccountData,
 } from "./types";
@@ -52,9 +53,19 @@ export async function supportProposal(
   const splVoteAccount = new PublicKey(validatorVoteAccount.voteAccount);
 
   // The handler re-tallies every existing supporter, so the compute budget is
-  // sized from the current list length rather than a flat worst case.
-  const proposalAccount = await program.account.proposal.fetch(proposalPubkey);
-  const numSupporters = Number(proposalAccount.numSupporters ?? 0);
+  // sized from the current list length rather than a flat worst case. If the
+  // read fails the support itself may still succeed, so fall back to the
+  // program's supporter cap rather than aborting on a budgeting detail.
+  let numSupporters = MAX_SUPPORTERS;
+  try {
+    const proposalAccount = await program.account.proposal.fetch(proposalPubkey);
+    numSupporters = proposalAccount.numSupporters;
+  } catch (error) {
+    console.warn(
+      `Could not read the supporter count for ${proposalId}; requesting the compute budget for a full ${MAX_SUPPORTERS}-supporter list`,
+      error,
+    );
+  }
 
   // Derive support PDA - based on IDL, it uses proposal and signer
   const supportPda = deriveSupportPda(
--- a/frontend/src/chain/instructions/supportProposal.ts
+++ b/frontend/src/chain/instructions/supportProposal.ts
@@ -11,7 +11,7 @@ import {
   SupportProposalParams,
   TransactionResult,
   SNAPSHOT_PROGRAM_ID,
-  SUPPORT_COMPUTE_UNIT_LIMIT,
+  supportComputeUnitLimit,
   ChainVoteAccountData,
 } from "./types";
 import {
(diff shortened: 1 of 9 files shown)


tiago18c:
Given the effort, would you mind making it scale linearly with the number of supporters?
Fetch the proposal state for the current number of supporters, use the 20k baseline plus ~132 per supporter and at the end add a 10-15% headroom.

frontend/src/chain/instructions/supportProposal.ts
```diff
@@ -45,6 +51,11 @@ export async function supportProposal(
   const proposalPubkey = new PublicKey(proposalId);
   const splVoteAccount = new PublicKey(validatorVoteAccount.voteAccount);
 
+  // The handler re-tallies every existing supporter, so the compute budget is
+  // sized from the current list length rather than a flat worst case.
+  const proposalAccount = await program.account.proposal.fetch(proposalPubkey);
+  const numSupporters = Number(proposalAccount.numSupporters ?? 0);
```
tiago18c: nit: if we can't retrieve the proposal we should fallback to the 2k supporters hard limit instead

tiago18c (changes requested):
just one more nit then up only!

tiago18c (approved):

PR #120, opened 2026-08-03
fix(ncn-router): Drop Verifiers That Have Fallen Behind The Fleet

0xIchigo:
## TLDR

This PR drops verifiers that have fallen behind the rest of the fleet

## Problem

`ncn-router` picks a verifier uniformly at random from those with `status == "ok"`. That status is set by the cron comparing each verifier's merkle root and snapshot hash against the on-chain ballot at the slot that verifier itself reported

An operator that stopped uploading months ago still passes: its stale data is perfectly self-consistent with the onchain ballot for its own old slot. So it stays `"ok"` forever, keeps its routing ticket, and 404s every proof request for a current snapshot. Staleness is never evaluated anywhere

## Measured Impact

Polled all 11 configured verifiers on 2026-08-03 (chain slot 437,011,715):

| Verifier | Slot | Behind |
|---|---|---|
| ncn-verifier.chainflow.io | 436941951 | current |
| verifier.promptlogic.systems | 436924558 | current |
| gov.lantern.one | 436919878 | current |
| verifier.titananalytics.io | 435942000 | ~5 days |
| verifier.nops.blocksize.dev | 422497000 | **~67 days** |
| ncn-verifier.digital-energy.io | 422497000 | **~67 days** |
| ncn.stakeware.xyz:3000 | 422497000 | **~67 days** |
| ncn.brewlabs.so | 422497000 | **~67 days** |
| ncn.ha1iad3.com | — | expired TLS cert |
| solgov.com | — | redirects https → http |
| ncn-verifier.exotechnologies.xyz:3000 | — | timeout |

Six samples of the public router endpoint returned digital-energy, brewlabs, stakeware, blocksize, and two timeouts; no current operator in six tries. A validator using the CLI's default `operator_api_url` has roughly a 1-in-4 chance
of reaching an operator that holds the slot they need

## Fix

1. Record each verifier's reported `slot` in the whitelist file. Also makes a lagging operator visible without re-polling
2. After the on-chain comparison, demote any `"ok"` verifier trailing the freshest whitelisted peer by more than `NCN_MAX_VERIFIER_SLOT_LAG` (default to one epoch) to `"stale"`, with the lag in `reason`

Applied to the fleet above: chainflow becomes `chosen_slot`; titan (999,951 behind) and the four frozen operators (14,444,951 behind) are demoted. The pool goes from 8 to 3, all of which can serve current proofs

## Tests
- `cargo test` in `ncn-router` → 27 passed (13 cron + 14 router), up from 19.
- Eight new cron tests cover the reported failure mode, the lag budget, the common-older-slot case, pool-never-empty, non-`ok` statuses being preserved, the no-`ok`-verifiers case, the inclusive boundary, and lag-override parsing (valid, whitespace, `0`-as-strict-policy, absent, and five malformed inputs)
- One new router test deserializes a whitelist containing the new `slot` field. The cron writes that file and the router reads it, so the two structs are a cross-binary contract; if the reader ever gained `deny_unknown_fields`, adding a field would make it fail to parse the whole whitelist and stop routing entirely. The test pins that

### DIFF
--- a/ncn-router/README.md
+++ b/ncn-router/README.md
@@ -35,6 +35,7 @@
   - `SOLANA_RPC_URL` (optional): Solana RPC endpoint (defaults to mainnet/testnet public RPCs)
   - `NCN_PROGRAM_ID` (required): NCN program ID on Solana
   - `NCN_WHITELIST_MAINNET_PATH` / `NCN_WHITELIST_TESTNET_PATH` (optional): Output paths for whitelist JSON (defaults to `ncn_whitelist.mainnet.json` / `ncn_whitelist.testnet.json`)
+  - `NCN_MAX_VERIFIER_SLOT_LAG` (optional): How many slots a verifier's snapshot may trail the freshest whitelisted verifier before it is marked `stale` and dropped from routing (default: `432000`, one epoch / ~2 days). A verifier that stopped uploading still matches the on-chain ballot for its own old slot, so it stays `ok` without this bound and 404s every current proof request.
 
 ### Usage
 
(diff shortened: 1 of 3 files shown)


MicaiahReid (approved):
LGTM! Thanks @0xIchigo!

PR #121, opened 2026-08-03
fix(frontend): Degrade Gracefully When Validator Metadata is Unavailable

0xIchigo:
## TLDR

This PR ensures we degrade gracefully whenever validator metadata is unavailable, so the governance UI doesn't silently zero out 

## Problem

`useGetValidators` fetched StakeWiz metadata and RPC vote accounts with `Promise.allSettled`, required **both** to resolve, and returned `[]` otherwise

`allSettled` never rejects, so react-query recorded a successful query returning an empty list. No error state, no retry, no signal, and every consumer computed a zero denominator from data that looked fine

Twelve files depend on this query. On a StakeWiz outage:

| Consumer | Renders |
|---|---|
| `VoteBreakdown` | every vote percentage **0%** |
| `support-phase-progress` | **0 SOL** total staked, threshold **0** |
| `DashboardStats` | **0 active validators** |
| `TopVotersTable` | every voter **0%** |
| `useValidatorsVoterSplits` | query disabled (gated on `length > 0`) |

During a live proposal that reads as "this proposal has no support," which is about the worst thing a governance UI can say incorrectly

## Fix

Separate the two sources by what each is actually for:

- RPC is the source of truth for the validator set and its stake, so a failure now throws. React-query retries, and consumers can distinguish "broken" from "zero"
- StakeWiz only supplies display metadata, so it is allowed to fail: validators fall back to placeholder names with correct stake numbers. This is the far more likely outage of the two, and it is now fully non-fatal

Note that all twelve consumers already guard against `undefined` (`?.length || 0`, gated `enabled`, or an explicit `undefined` throw), so the new error path breaks nothing

## Tests

Four new tests:
- StakeWiz down → stake stays correct, only names degrade
- RPC down → query error rather than an empty list
- Both resolve with conflicting stake → the RPC value wins
- Metadata omits `identity` → filled from the RPC's `nodePubkey`

### DIFF
--- a/frontend/src/components/governance/shared/DashboardStats.tsx
+++ b/frontend/src/components/governance/shared/DashboardStats.tsx
@@ -59,7 +59,9 @@ export function DashboardStats({
 
   const totalStaked =
     stakeAccounts?.reduce((acc, curr) => acc + curr.activeStake, 0) || 0;
-  const activeValidators = validators?.length || 0;
+  // Left undefined when the validator query has not resolved; formatOptionalCount
+  // renders "-" rather than a "0" that reads as a real count.
+  const activeValidators = validators?.length;
 
   // TODO: on mobile only show "custom" when the network is a custom URL, otherwise show the network name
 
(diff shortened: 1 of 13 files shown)


tiago18c:
Thank you for this! 
Would you mind rebasing to the latest main and doing the same for the new supporters pane as well?

tiago18c (approved):

PR #122, opened 2026-08-03
fix(frontend): Read the Support Threshold From the Onchain Config

0xIchigo:
## TLDR

The support threshold was hardcoded to 10% while mainnet requires 15%, so the progress bar reported every proposal at roughly 1.5x its real progress

Reviewing that turned up two more live misreports in the same panel: the support percentage was 10x too high, and the "threshold reached" banner fired on every page load before the data arrived

### 1. Threshold hardcoded to 10%

```ts
// TODO: testnet is 10%, mainnet is 15% - create a getter depending on the endpoint type
export const SUPPORT_THRESHOLD_PERCENT = 10;
```

Mainnet's `cluster_support_pct_min_bps` is `1500` (15%, set in #99), so `requiredThresholdLamports = totalStakedLamports * 0.10` understated the bar by a third. With 12% of stake supporting a proposal:

| threshold used | outcome shown |
|---|---|
| 10% (hardcoded) | **advances to discussion** |
| 15% (actual chain value) | **fails** |

It also leaked into copy: `LifecycleIndicator` and `SupportProposalSection` both told validators the phase "requires 10% of total validator stake."

### 2. Support percentage 10x too high

`supportPercentOfTotal` was multiplied by `1000` instead of `100`. A proposal with 1.2% of stake rendered a "12.00%" badge, and the same value drove a progress bar that was ten times too full

### 3. "Threshold reached" banner on first paint

`isThresholdMet` was `currentSupportLamports >= requiredThresholdLamports`, and `requiredThresholdLamports` is `0` until validator stake and the governance config load, so any support satisfied it. The banner is not gated by `isLoading` (that only gates individual stat elements), so every support-phase page load briefly displayed "Support threshold reached! Proposal advancing to next phase."

## Fix

The threshold value was already in reach; `GovernanceConfigDto` carries `clusterSupportPctMinBps`, and `SupportPhaseProgress` already consumed the governance config for epoch lengths and used it only for those

In this PR, we:

- Add `supportThresholdPercentFromGovernanceConfig()` beside the existing `epochConstantsFromGovernanceConfig()`, so both lifecycle parameters derive from the config the same way.
- Thread the percentage through `getProposalStatus` as a parameter. It is a pure function, so a parameter is the right shape, and it removes a `lib/` → component import that forced the existing tests to `jest.mock` a UI module in order to test status logic
- The two components that baked the number into module-level constants now build that copy per render from the config context, rendering `—` while it loads rather than a wrong number
- Extract the panel's figures into a pure `computeSupportStats()`. Three display-correctness bugs in one component with no coverage on two of them is how a fourth arrives. The extraction is a mechanical move; the only logic changes are `1000` → `100` and requiring a positive threshold before `isThresholdMet` can be true

`grep -rn "SUPPORT_THRESHOLD_PERCENT" src` returns nothing. No per-cluster getter is needed; the value follows whichever cluster the app is pointed at

## Tests
- `pnpm test` → 86 passed, 1 skipped (up from 76)
- `getProposalStatus`; 12% support advances under a 10% threshold and fails under 15%, pinning the misreport in both directions. Plus bps conversion (1500 → 15, 1000 → 10, 50 → 0.5). The existing 34 status tests now pass the threshold explicitly instead of relying on a mocked module constant
- `computeSupportStats` has seven cases: percentage of total stake (the 10x regression), progress against the configured threshold, exact-threshold equality, no success before the threshold is known, no success before stake is known, no division by a zero validator count, and progress exceeding 100%

## Verified

- All three `LifecycleIndicator` call sites sit under the app-level `GovernanceConfigProvider`, and no test renders it, so the new hook call cannot throw
- `tsc --noEmit` clean, eslint clean on every touched file

### DIFF
--- a/frontend/src/components/proposals/detail/ProposalDetailView.tsx
+++ b/frontend/src/components/proposals/detail/ProposalDetailView.tsx
@@ -6,6 +6,7 @@ import ProposalDetailHeader from "./ProposalDetailHeader";
 import VoteBreakdown from "./VoteBreakdown";
 import PhaseTimeline from "./phase-timeline";
 import TopVotersTable from "./TopVotersTable";
+import TopSupportersTable from "./TopSupportersTable";
 import CastVoteWrapper from "./CastVote";
 import SupportProposalSection from "./SupportProposalSection";
 import { useWallet } from "@solana/wallet-adapter-react";
@@ -107,7 +108,13 @@ export default function ProposalDetailView({
         </>
       )}
       <PhaseTimeline proposal={proposal} isLoading={isLoading} />
-      <TopVotersTable proposal={proposal} />
+      {/* Before voting activates the proposal has no votes, only support
+          pledges — show the supporters instead of an empty voters table. */}
+      {isSupportPhaseView ? (
+        <TopSupportersTable proposal={proposal} />
+      ) : (
+        <TopVotersTable proposal={proposal} />
+      )}
     </div>
   );
 }
--- a/frontend/src/components/proposals/detail/TopSupportersColumns.tsx
+++ b/frontend/src/components/proposals/detail/TopSupportersColumns.tsx
@@ -1,0 +1,77 @@
+"use client";
+
+import { ColumnDef } from "@tanstack/react-table";
+import { SortableHeaderButton } from "@/components/governance/shared/SortableHeaderButton";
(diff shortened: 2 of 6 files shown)


0xIchigo:
My repo was based wasn't on the latest commit that fixed this 😭

PR #123, opened 2026-08-03
fix(frontend): Don't Report Support Success Before Stake Loads

0xIchigo:
## TLDR

This PR is a follow-up to #118. Two zero-state bugs survive in the same panel: the support page claims "Support threshold reached!" on every first paint, and participation renders `NaN%` until the validator query resolves

## Issues

#118 correctly sourced the threshold from the on-chain config with a 15% fallback, which closed the config route into the first bug. The validator-stake route is still open:

```ts
const isThresholdMet = currentSupportLamports >= requiredThresholdLamports;
```

Validator stake arrives from a separate query, so `requiredThresholdLamports` is `0` on first paint and any support satisfies `>= 0`. The banner is not gated by `isLoading`—that only gates individual stat elements—so every support-phase page load briefly displays "Support threshold reached! Proposal advancing to next phase."

```ts
const participationPercent = (validatorCount / numOfValidators) * 100;
```

`NaN` whenever `numOfValidators` is 0, rendered directly as "NaN%".

## Fix

Extract the panel's figures into a pure `computeSupportStats()` and guard both zero states. The extraction is a faithful move of the block #118 edited; it keeps `supportThresholdPercentFromConfig` and the `configData` memo dependency as introduced there. The only logic changes are the two guards

## Tests
- `pnpm test` → 82 passed, 1 skipped
- `tsc --noEmit` and eslint clean

### DIFF
--- a/frontend/src/components/proposals/detail/support-phase-progress/SupportDonut.tsx
+++ b/frontend/src/components/proposals/detail/support-phase-progress/SupportDonut.tsx
@@ -5,32 +5,26 @@ import { useSpring, animated, to } from "@react-spring/web";
 import { formatSOL } from "@/lib/governance/formatters";
 import { DONUT_CONFIG, DonutChartBase } from "../shared/DonutChartBase";
 
+/**
+ * Presentational only. These figures come from `computeSupportStats` so the
+ * donut and its parent cannot disagree — the donut used to recompute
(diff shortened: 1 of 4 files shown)


tiago18c (approved):
Thank you !

PR #126, opened 2026-08-04
docs(svmgov): add clone + release tag steps to quick setup

lostintime101:
This PR implements a change based on [validator feedback](https://discord.com/channels/428295358100013066/673718028323782674/1534123634808066081) in the Solana Tech Discord mb-planning channel

Both Quick Setups started at `cd svmgov/cli && ./install.sh`, which assumes a repo that's already cloned and says nothing about which ref to build. Add the clone and checkout steps to the validator and staker pages, and explain the tag convention.

The checkout resolves the newest release tag at run time rather than naming a version, so the pages can't go stale as releases land:

    git checkout "$(git describe --tags --abbrev=0 --match 'v*')"

`--match 'v*'` skips the stray non-release `list` tag. Tags are `v<version>`, match the version in svmgov/cli/Cargo.toml and the program crate, and pushing one triggers .github/workflows/release.yaml, so they're real releases rather than bookmarks. main routinely sits ahead of the latest tag, so "just build main" would silently drift.

### DIFF
--- a/docs/src/content/svmgov/stakers/index.mdx
+++ b/docs/src/content/svmgov/stakers/index.mdx
@@ -45,7 +45,12 @@ The Merkle proof system ensures that:
 ## Quick Setup
 
 ```bash
-# Install svmgov
+# Clone the repo and check out the latest release tag
+git clone https://github.com/solana-foundation/solana-governance.git
+cd solana-governance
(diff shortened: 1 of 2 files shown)


tiago18c (approved):
Thanks, this is great.

We're also planning on publishing crates so that we can support `cargo install svmgov-cli` directly.

PR #131, opened 2026-08-04
fix(frontend): Discussion phase calculation is innacurate

tigarcia:
The discussion phase was not being calculated correctly due to the variable lenght of the support phase.

### DIFF
--- a/frontend/src/lib/proposals.ts
+++ b/frontend/src/lib/proposals.ts
@@ -76,7 +76,6 @@ export interface ProposalPhaseEpochs {
 export interface ProposalPhaseAnchors {
   voting: boolean;
   startEpoch: number;
-  endEpoch: number;
 }
 
 export function getProposalPhaseEpochs(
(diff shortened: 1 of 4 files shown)


frontend/src/lib/proposals.ts
```diff
@@ -66,9 +66,23 @@ export interface ProposalPhaseEpochs {
...
+ * On-chain phase anchors from the proposal account. Once support succeeds
+ * (`voting === true`) the program records the definitive voting start/end
+ * epochs, which supersede any creation-based projection: a proposal that
+ * reaches the threshold before its full support window enters discussion
+ * (and voting) earlier than the worst-case estimate.
+ */
+export interface ProposalPhaseAnchors {
+  voting: boolean;
+  startEpoch: number;
+  endEpoch: number;
```
MicaiahReid: Doesn't look like this value is being used anywhere. Is that intentional?

tigarcia: Not intentional. I removed it.

MicaiahReid (approved):
LGTM!

PR #132, opened 2026-08-04
fix(fe/svmgov): proposal description validation

MicaiahReid:
- Fix frontend to display proposal information when a PR link is used as the on-chain description
- Add validation in frontend proposal creation and svmgov CLI proposal creation so that:
  - proposal link meets on-chain requirements (github link, size requirements, etc)
  - link is to a valid markdown file, not a PR


tiago18c (approved):
lgtm!

PR #142, opened 2026-08-05
feat(svmgov): add support threshold on proposal acc for frontend support

tiago18c:
## Problem

There is a slight drift between onchain `get_epoch_total_stake` and offchain sum of all active stake. `get_epoch_total_stake` reports active stake at the end of epoch without rewards, as the offchain calculations report the reward values as well.

## Solution

Track the threshold and epoch for each proposal support for proper front end reporting.

## Changes

- Add `support_threshold` on proposal account, tracks expected threshold for the support phase 
- Add `last_support_epoch` on proposal account, tracks epoch boundaries to support recalculating `support_threshold`
- Retally on support only happens if we cross a epoch boundary from last vote

## Todo

Need to discuss right migration path, or if its even worth the effort.

### DIFF
--- a/svmgov/program/programs/svmgov_program/src/instructions/create_proposal.rs
+++ b/svmgov/program/programs/svmgov_program/src/instructions/create_proposal.rs
@@ -15,7 +15,7 @@ use crate::{
     events::ProposalCreated,
     stake_weight_bp,
     state::{GlobalConfig, Proposal, ProposalIndex},
-    utils::is_valid_github_link,
+    utils::{is_valid_github_link, min_stake_threshold},
 };
 
 #[derive(Accounts)]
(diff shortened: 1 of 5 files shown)


tiago18c:
Possible migration path:
- add the new fields after `num_supporters`
- add migration instruction that loads the account using the old layout (or even the new one as all proposals have supporters, so the new fields would deserialize from the supporters vec, but we don't care about those values right now)
- calculate expected account size from `INIT_SIZE + num_supporters * 32 + 8` and compare it against current account size to check if migration is required or already executed for the given account
- if migration is required
  - resize the account to add the new 16 bytes
  - reload the account using the new layout
  - set the new values

Issue #141, opened 2026-08-05
CLI mislabels voting-activated proposals before start_epoch as support and surfaces misleading ProposalClosed error


solanaqubits:
## Summary

`svmgov list-proposals --status support` can return a proposal whose on-chain `voting` flag is already `true`, while `svmgov support-proposal` then fails with `ProposalClosed: Proposal voting period has ended`.

For the observed proposals, voting had been activated but the formal voting window had not started. Both the CLI status and the surfaced on-chain error message are therefore misleading.

## Environment

- Repository: [solana-foundation/solana-governance](https://github.com/solana-foundation/solana-governance)
- CLI: `svmgov 0.4.0-40100`
- Source commit: [`a83de0a34feb646a3daf54c8054b16b06a6282f8`](https://github.com/solana-foundation/solana-governance/commit/a83de0a34feb646a3daf54c8054b16b06a6282f8)
- Public RPC endpoint: `https://api.mainnet-beta.solana.com`
- Current epoch when observed: 1012

## Affected public proposals

- `7QJD8MzheHWJLHS39NkoAbFCGFKg5d9QbVviRqD4YExP` — Double Disinflation
- `AGHDQ6gjRFJPoyEcHuc4X7sbxJwyJfeKTb3UrGFzFNZD` — SGP-0003: Resource and Inclusion Fee

## Reproduction and observed behavior

Both proposals were returned by:

```bash
svmgov --rpc-url https://api.mainnet-beta.solana.com list-proposals --status support --json
```

At epoch 1012, the JSON output reported `"status": "support"`, while also reporting `"voting": true`, `"start_epoch": 1021`, and `"end_epoch": 1024`.

Attempting `support-proposal` for the first proposal failed during simulation, before broadcast. The relevant error fields were:

```text
AnchorError ... support_proposal.rs:88
Error Code: ProposalClosed (6008)
Error Message: Proposal voting period has ended
```

No transaction was broadcast.

## Cause

[`get_proposal_status()`](https://github.com/solana-foundation/solana-governance/blob/a83de0a34feb646a3daf54c8054b16b06a6282f8/svmgov/cli/src/utils/commands.rs#L264-L293) returns `voting` only when `proposal.voting == true` and `current_epoch >= start_epoch`. For these activated proposals, `current_epoch` was still below `start_epoch`, so the function fell through to the pre-voting epoch arithmetic and returned `support`.

[...]

PR #144, opened 2026-08-05
fix(svmgov-cli): Derive Proposal Phase From Onchain State Instead of Creation Epoch

0xIchigo:
## TLDR

`svmgov proposal <id>` reported the double disinflation proposal as still in the support phase after it had already advanced, alongside a phase timeline whose voting window began before its own snapshot. The CLI recomputed the schedule from
`creation_epoch` plus config durations instead of reading the epochs the program had already committed onchain

The frontend half of this was fixed in #133; this is the CLI half

Note that this closes part of #141

## Cause

`activate_voting` anchors a proposal's schedule on whichever epoch support crossed the threshold, which can be any epoch inside the support window. The CLI's `get_proposal_status` consulted `voting` only for the voting and ended
phases, then fell through to a model built from `creation_epoch + config`:

```rust
if proposal.voting && current_epoch >= proposal.end_epoch { return "ended"; }
if proposal.voting && current_epoch >= proposal.start_epoch { return "voting"; }

// falls through here even when voting is true
let support_end = proposal.creation_epoch + config.max_support_epochs;
if current_epoch <= support_end { "support" } ...
```

That model is only correct for a proposal that crosses on the very last epoch of its window

There are three distinct bugs, with one cause:

1. **Status wrong** for any proposal that crosses early
2. **Timeline self-contradictory** — real on-chain epochs printed alongside projected ones, with no indication which was which
3. **`Failed` unreachable** — an unactivated proposal past its support window reported `support` indefinitely, though `support_proposal` and `retally_support` both reject past `creation_epoch + max_support_epochs`

All four CLI surfaces (i.e., detail view, `--status` filter, JSON output, list table) share `get_proposal_status`, so all four were affected

## Fix

We fix this with a new pure `svmgov/cli/src/utils/phase.rs`. Once `voting` is set, the onchain epochs are authoritative, and the config model is never consulted; it is used only to project a schedule for an unactivated proposal, labeled
`(projected — set once support crosses)`

The epoch support crossed the threshold is **not recorded on-chain**, and cannot be recovered from `start_epoch`:

- `flush_merkle_root` re-anchors with `target = current_epoch + snapshot_epoch_extension`, omitting `discussion_epochs`, so inverting `activate_voting`'s formula against a recovered proposal can produce a crossing epoch that predates the proposal; a recovery at epoch 1015 sets `start_epoch = 1017`, and `1017 - 1 - 7 - 1 = 1008` against a `creation_epoch` of 1011
- `update_config` can change `discussion_epochs` and `snapshot_epoch_extension` after activation, so the stored `start_epoch` may have been computed with values the inversion no longer sees

So it is reported as unknown rather than guessed. `support` and the discussion start are `Option`, and an activated proposal prints only what is stored:

```
created: epoch 1011 | discussion: through epoch 1019 |
snapshot: epoch 1020 | voting: epochs 1021-1024
```

The snapshot epoch comes from `start_epoch - 1`, which holds by construction in both the `activate_voting` and `flush_merkle_root` paths; more robust than dividing `snapshot_slot`, which depends on `snapshot_slot_offset`. For this proposal it gives 1020, matching the stored `snapshot_slot` of 440,641,000 (1020 × 432,000 + 1,000)

## Tests

Twelve tests in the new module, built on the real proposal's onchain values rather than invented ones. `GlobalConfig` was read from `4V8PnDoVa7BSUWj16uhUkgHhXPoBEvMSk4EFhKA9wyFA` and the proposal from its own account, so the fixture is not inferred from the CLI output being fixed

The ones carrying the most weight:

- **`an_early_crossing_is_not_reported_as_still_gathering_support`** — the reported case: epoch 1012 reports Discussion, and not Support
- **`the_timeline_agrees_with_the_phase_at_every_epoch`** — walks every epoch of the timeline and checks `proposal_phase` reports the phase whose window it falls in. An ordering-only check passed while the two definitions disagreed about a boundary epoch, so this replaces it
- **`a_flushed_proposal_does_not_get_a_fabricated_crossing_epoch`** — a `flush_merkle_root`-recovered proposal (`start_epoch = 1017`), asserting nothing is claimed about the crossing and the rest of the timeline stays correct
- **`changing_the_config_durations_cannot_corrupt_an_activated_timeline`** — the same proposal with very different durations must yield an identical timeline
- **`an_unactivated_proposal_past_its_window_has_failed`** — previously unreachable

`cargo test` in `svmgov/cli` → 42 passed. Build clean, `phase.rs` rustfmt-clean as well

### DIFF
--- a/svmgov/cli/src/utils/commands.rs
+++ b/svmgov/cli/src/utils/commands.rs
@@ -15,6 +15,7 @@ use serde::{Deserialize, Serialize};
 use crate::{
     anchor_client_setup,
     svmgov_program::accounts::{GlobalConfig, Proposal},
+    utils::phase::{epochs_remaining, phase_timeline, proposal_phase, PhaseInputs},
     utils::utils::fetch_global_config,
 };
 
@@ -102,21 +103,10 @@ fn print_proposal_detail(proposal_id: &str, proposal: &Proposal, current_epoch: 
     let cluster_support_sol = proposal.cluster_support_lamports as f64 / 1_000_000_000.0;
     let proposer_stake_bp = proposal.proposer_stake_weight_bp as f64 / 100.0;
 
-    let status = get_proposal_status(proposal, current_epoch, config);
-    let status_display = match status {
-        "support" => "Support",
-        "discussion" => "Discussion",
-        "snapshot" => "Snapshot",
-        "voting" => "Voting",
-        "ended" => "Ended (awaiting finalization)",
-        "finalized" => "Finalized",
-        other => other,
-    };
-
-    // Compute phase boundaries
-    let support_end = proposal.creation_epoch + config.max_support_epochs;
-    let discussion_end = support_end + config.discussion_epochs;
-    let snapshot_end = discussion_end + config.snapshot_epoch_extension;
+    let inputs = phase_inputs(proposal, config);
+    let phase = proposal_phase(&inputs, current_epoch);
+    let status_display = phase.label();
+    let timeline = phase_timeline(&inputs);
 
     table.add_row(vec![Cell::new("Proposal ID"), Cell::new(proposal_id)]);
     table.add_row(vec![Cell::new("Title"), Cell::new(&proposal.title)]);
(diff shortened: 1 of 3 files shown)


svmgov/cli/src/utils/phase.rs
```diff
@@ -0,0 +1,390 @@
...
+//! durations therefore only matches reality for a proposal that crosses on the
+//! very last epoch of its window — for anything that crosses early it reports
+//! the proposal as still gathering support long after it has advanced.
+//!
+//! So: once `voting` is set, the on-chain epochs are authoritative and the
+//! config model is not consulted. It is used only to project a schedule for a
+//! proposal that has not yet activated.
+
+/// Fields needed to place a proposal in its lifecycle. Plain values rather than
+/// the generated account types, so the logic is unit-testable.
```
MicaiahReid: This reads more like how this mod fixes the previous error, rather than what the module actually does for us. Could this be reworded to instead describe the purpose of the mod, how the purpose is acheived, and high level inputs/outputs?

svmgov/cli/src/utils/phase.rs
```diff
@@ -0,0 +1,390 @@
...
+            Self::Snapshot => "snapshot",
+            Self::Voting => "voting",
+            Self::Ended => "ended",
+            Self::Finalized => "finalized",
+            Self::Failed => "failed",
+        }
+    }
+}
+
+pub fn proposal_phase(p: &PhaseInputs, current_epoch: u64) -> ProposalPhase {
```
MicaiahReid: IMO a cleaner, more standard pattern would for this to be an 
```rust
impl ProposalPhase {
  pub fn new(p: &PhaseInputs, current_epoch: u64) -> Self {
  ...
  }
}
```
rather than a separate function

svmgov/cli/src/utils/phase.rs
```diff
@@ -0,0 +1,390 @@
...
+    /// `None` once activated — see the note above.
+    pub support: Option<(u64, u64)>,
+    /// Start is `None` once activated; the end is always known.
+    pub discussion: (Option<u64>, u64),
+    pub snapshot: u64,
+    pub voting: (u64, u64),
+    pub projected: bool,
+}
+
+pub fn phase_timeline(p: &PhaseInputs) -> PhaseTimeline {
```
MicaiahReid: Same note as above, this should be a `new` function on the `PhaseTimeline` struct

svmgov/cli/src/utils/commands.rs
```diff
@@ -261,36 +252,23 @@ struct ProposalOutput {
...
-
-    // Once start_epoch/end_epoch are set (voting flag raised during support),
-    // use those to determine voting/ended phases
-    if proposal.voting && current_epoch >= proposal.end_epoch {
-        return "ended";
-    }
-    if proposal.voting && current_epoch >= proposal.start_epoch {
-        return "voting";
+/// Bridges the generated account types into the pure phase logic.
+fn phase_inputs(proposal: &Proposal, config: &GlobalConfig) -> PhaseInputs {
```
MicaiahReid: I believe this should be a `new` function on `PhaseInputs`

MicaiahReid (changes requested):
Awesome stuff! I've got some cleanup/code style requests

MicaiahReid (approved):
Looks good, thanks @0xIchigo! Will merge once github actions are online again :/

PR #145, opened 2026-08-05
fix(svmgov): Distinguish support-already-activated from voting-ended

0xIchigo:
## TLDR

A validator whose `support-proposal` was rejected because the proposal had already been activated was told "Proposal voting period has ended" when voting had not yet started. This gives that state its own error, and makes the CLI refuse before building a transaction the program will reject

## Scope 

The issue raises three problems. Two are already fixed in #144 (in review):

| Ask | Where |
|---|---|
| `list-proposals` must not label `voting == true` as `support` | #144 |
| Activated-before-`start_epoch` needs an unambiguous non-support status | #144 reports Discussion, or Snapshot in the epoch before `start_epoch` |
| `ProposalClosed` message is misleading | this PR |

The reporter's root-cause analysis of `get_proposal_status` is exactly right,
and matches what we hit independently on Double Disinflation.

## Bug

`support_proposal` and `retally_support` both gated on:

```rust
require!(!voting && !finalized, GovernanceError::ProposalClosed);
```

One error covering two unrelated states, with a message that is wrong for both. `activate_voting` sets `voting` the instant support crosses the threshold, while `start_epoch` is still epochs away, so "the voting period has ended" describes
the opposite of what happened

## Fix

```rust
require!(!self.proposal.finalized, GovernanceError::ProposalFinalized);
require!(!self.proposal.voting, GovernanceError::SupportAlreadyActivated);
```

`SupportAlreadyActivated`: "Support is closed: this proposal already reached its support threshold and voting has been scheduled"*

`ProposalClosed` is deliberately untouched in `cast_vote`, `cast_vote_override`, `modify_vote` and `modify_vote_override`, where it guards `current_epoch < end_epoch` and the existing message is accurate

### Error Code Stability

The variant is appended instead of inserted since Anchor derives error codes from declaration order. `ProposalClosed` remains 6008 (i.e., the code quoted in the issue), the new variant is 6058, and both IDL diffs are purely additive

### CLI guard

The issue's practical complaint is that validators are prompted to submit an impossible transaction. Since #119 the CLI already fetches the proposal to size the compute budget, so it can refuse up front and name the next action:

```
Proposal 7QJD8Mzhe... has already reached its support threshold, so support is
closed. Voting runs from epoch 1021 to 1024 — use `svmgov cast-vote` once epoch
1021 begins.
```

This is the half that lands without a program redeploy

## Tests

- `support_and_retally_reject_after_voting_activated` now asserts the new error, and additionally asserts the proposal is genuinely in the transition state the issue describes; `voting && !finalized && current_epoch < start_epoch`. Without
that the test would pass for the wrong reason, since any post-activation state produces the same error
- Full program suite green (11 targets), CLI 30 tests, frontend 270 tests
- `tsc` clean
- No new rustfmt drift

### DIFF
--- a/svmgov/program/programs/svmgov_program/src/instructions/support_proposal.rs
+++ b/svmgov/program/programs/svmgov_program/src/instructions/support_proposal.rs
@@ -84,10 +84,15 @@ impl<'info> SupportProposal<'info> {
     pub fn support_proposal(&mut self, bumps: &SupportProposalBumps) -> Result<()> {
         let clock = Clock::get()?;
 
-        // Ensure proposal is eligible for support
+        // Ensure proposal is eligible for support. Split from a single
+        // ProposalClosed check so the two states are distinguishable: an
+        // activated proposal has *not* finished voting — voting may not even
+        // have started — and "the voting period has ended" sent validators
+        // looking for the wrong problem.
+        require!(!self.proposal.finalized, GovernanceError::ProposalFinalized);
         require!(
-            self.proposal.voting == false && self.proposal.finalized == false,
-            GovernanceError::ProposalClosed
+            !self.proposal.voting,
+            GovernanceError::SupportAlreadyActivated
         );
 
         // Support is accepted anywhere inside the window
--- a/svmgov/program/programs/svmgov_program/src/instructions/support_proposal.rs
+++ b/svmgov/program/programs/svmgov_program/src/instructions/support_proposal.rs
@@ -84,11 +84,7 @@ impl<'info> SupportProposal<'info> {
     pub fn support_proposal(&mut self, bumps: &SupportProposalBumps) -> Result<()> {
         let clock = Clock::get()?;
 
-        // Ensure proposal is eligible for support. Split from a single
-        // ProposalClosed check so the two states are distinguishable: an
-        // activated proposal has *not* finished voting — voting may not even
-        // have started — and "the voting period has ended" sent validators
-        // looking for the wrong problem.
+        // A proposal that is already finalized or being voted on is not eligible for support
         require!(!self.proposal.finalized, GovernanceError::ProposalFinalized);
         require!(
             !self.proposal.voting,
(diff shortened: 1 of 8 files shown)


svmgov/program/programs/svmgov_program/src/instructions/support_proposal.rs
```diff
@@ -84,10 +84,15 @@ impl<'info> SupportProposal<'info> {
...
+        // ProposalClosed check so the two states are distinguishable: an
+        // activated proposal has *not* finished voting — voting may not even
+        // have started — and "the voting period has ended" sent validators
+        // looking for the wrong problem.
+        require!(!self.proposal.finalized, GovernanceError::ProposalFinalized);
         require!(
-            self.proposal.voting == false && self.proposal.finalized == false,
-            GovernanceError::ProposalClosed
+            !self.proposal.voting,
+            GovernanceError::SupportAlreadyActivated
```
MicaiahReid: IMO this comment is keeping track of legacy stuff that's being removed - no need to say how the code used to be. 

Ideally previously fixed code would be captured in two regression tests - one to verify finalized state yields `ProposalFinalized` error, the other to verify voting state yields `SupportAlreadyActivated`.

```suggestion
        // A proposal that is already finalized or being voted on is not eligible for support
        require!(!self.proposal.finalized, GovernanceError::ProposalFinalized);
require!(
            !self.proposal.voting,
            GovernanceError::SupportAlreadyActivated
```

svmgov/program/programs/svmgov_program/tests/support_150_of_1000.rs
```diff
@@ -119,8 +119,14 @@ fn support_proposal_remeasures_prior_stake_across_epochs() {
...
+/// Once voting is active, both `support_proposal` and `retally_support` reject
+/// with `SupportAlreadyActivated`.
+///
+/// The error is deliberately not `ProposalClosed` ("the voting period has
+/// ended"): activation happens the moment the threshold is crossed, while
+/// `start_epoch` is still in the future, so at this point voting has not even
+/// begun. Reporting it as ended sent validators looking for the wrong problem
+/// (issue #141).
 #[test_log::test]
 fn support_and_retally_reject_after_voting_activated() {
```
MicaiahReid, line 131: Could we augment this test or add a new one to also test the `ProposalFinalized` error is found when in finalized state?

MicaiahReid (changes requested):

MicaiahReid (approved):
Looks good!

tiago18c (approved):
lgtm!

PR #148, opened 2026-08-07
fix(svmgov-cli): Harden the Vote Path Against Operator API Failures

0xIchigo:
## TL;DR

I swept the CLI vote path ahead of voting opening at epoch 1021 and found four ways it can fail badly: 
- an unbounded request that can hang forever
- an opaque error for the 404 a stale operator returns
- a panic on malformed responses, and
- a leaf/PDA inconsistency that would fail with no explanation

All are on `cast-vote` and `cast-vote-override`, which are the commands validators run to vote

## 1. No request timeout

The proof fetches used `reqwest::get`, whose default client has no timeout. `proposal_link.rs` already builds its own client for precisely this reason, with the comment "the default client has no timeout"; the proof path never got the same treatment

Both fetches are now bounded at 15s

## 2. Unhandled non-200

A non-success response went straight into `response.json()`, so the 404 a stale operator returns for a snapshot it has not uploaded surfaced as an opaque deserialization failure. With four operators still frozen at slot 422497000, that 404 is an expected outcome rather than an edge case

Now:

```
The operator API at https://... has no vote account proof for this snapshot slot
(404). It may not have uploaded this snapshot yet — retry, or point
--operator-api-url at another operator.
```

## 3. Leaf/PDA inconsistency

`init_meta_merkle_proof` seeds its PDA from `meta_merkle_leaf.vote_account`. The old code built that leaf from the locally derived vote account while passing a PDA derived from the API's. Identical in the normal case, but any divergence produced an Anchor `ConstraintSeeds` failure with nothing pointing at the cause

Sourcing the whole leaf from the response makes them consistent. Flagging it explicitly because it is a behavior change sitting inside what otherwise reads as a pure refactor

That also means the CLI now takes the account identity on trust, so it is now checked: a proof returned for a different vote or stake account than the one requested fails immediately, naming both

## 4. Panic on a malformed response

`Pubkey::from_str_const` decodes base58 in a const context and panics on bad input:

```
""                                 -> PANIC
"abc"                              -> PANIC
"not-base58-0OIl"                  -> PANIC
"11111111111111111111111111111111" -> ok
```

Both vote paths passed `stake_merkle_root` and every proof hash from the operator response straight into it, so a malformed reply aborted the CLI with a stack trace instead of an error a validator could act on

Fixed by using `MetaMerkleLeaf::try_from` and `convert_merkle_proof_strings`, validating converters that already existed in the same module and were already used on the stake-proof path. The vote path simply wasn't using them

## Tests

- Six tests covering every input that previously panicked — empty, short, non-base58, and valid-base58-but-wrong-length — asserting an `Err` rather than an abort, plus the valid cases, the legitimately empty proof a single-leaf tree produces, and the wrong-account rejection
- `grep` confirms no `from_str_const` on runtime data and no bare `reqwest::get` remains anywhere in the CLI
- `cargo test` in `svmgov/cli` → 36 passed
- Build clean with no new warnings
- `api_helpers.rs` at its rustfmt baseline

### DIFF
--- a/svmgov/cli/src/instructions/cast_vote.rs
+++ b/svmgov/cli/src/instructions/cast_vote.rs
@@ -10,7 +10,7 @@ use crate::{
     constants::*,
     svmgov_program::{accounts::Proposal, client::{accounts, args}},
     utils::{
-        api_helpers::{self, get_vote_account_proof},
+        api_helpers::{self, convert_merkle_proof_strings, get_vote_account_proof},
         utils::{
             compute_vote_expiry_timestamp, create_spinner, derive_vote_override_cache_pda,
             derive_vote_pda, setup_all,
@@ -92,8 +92,11 @@ pub async fn cast_vote(
 
         let init_spinner = create_spinner("Initializing meta merkle proof...");
 
-        let voting_wallet = Pubkey::from_str(&proof_response.meta_merkle_leaf.voting_wallet)
-            .map_err(|e| anyhow!("Invalid voting wallet in proof: {}", e))?;
+        // Validate every field from the operator API rather than assuming it is
+        // well-formed: these converters return a clear error, where the const
+        // base58 decoder would panic on a malformed response.
+        let meta_merkle_leaf = MetaMerkleLeaf::try_from(&proof_response.meta_merkle_leaf)?;
+        let meta_merkle_proof = convert_merkle_proof_strings(&proof_response.meta_merkle_proof)?;
 
         let close_timestamp = match close_timestamp_override {
             Some(ts) => ts,
@@ -105,20 +108,8 @@ pub async fn cast_vote(
             .request()
             .args(ncn_snapshot::instruction::InitMetaMerkleProof {
                 close_timestamp,
-                meta_merkle_leaf: MetaMerkleLeaf {
-                    voting_wallet,
-                    vote_account,
-                    stake_merkle_root: Pubkey::from_str_const(
-                        proof_response.meta_merkle_leaf.stake_merkle_root.as_str(),
-                    )
-                    .to_bytes(),
-                    active_stake: proof_response.meta_merkle_leaf.active_stake,
-                },
-                meta_merkle_proof: proof_response
-                    .meta_merkle_proof
-                    .iter()
-                    .map(|s| Pubkey::from_str_const(s).to_bytes())
-                    .collect(),
+                meta_merkle_leaf,
+                meta_merkle_proof,
             })
             .accounts(ncn_snapshot::accounts::InitMetaMerkleProof {
                 consensus_result: consensus_result_pda,
--- a/svmgov/cli/src/instructions/cast_vote.rs
+++ b/svmgov/cli/src/instructions/cast_vote.rs
@@ -93,8 +93,7 @@ pub async fn cast_vote(
         let init_spinner = create_spinner("Initializing meta merkle proof...");
 
         // Validate every field from the operator API rather than assuming it is
-        // well-formed: these converters return a clear error, where the const
-        // base58 decoder would panic on a malformed response.
+        // well-formed
         let meta_merkle_leaf = MetaMerkleLeaf::try_from(&proof_response.meta_merkle_leaf)?;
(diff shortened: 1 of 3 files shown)


0xIchigo:
can I get a review here pls 🥺 @MicaiahReid @tiago18c

MicaiahReid:
> can I get a review here pls 🥺 @MicaiahReid @tiago18c  

My bad! Will review today!

svmgov/cli/src/utils/api_helpers.rs
```diff
@@ -1,11 +1,66 @@
...
 use anchor_lang::prelude::Pubkey;
 use anyhow::{anyhow, Result};
 use log::info;
 use ncn_snapshot::{MetaMerkleLeaf, MetaMerkleProof, StakeMerkleLeaf};
 use serde::{Deserialize, Serialize};
 
+/// Bound on a single proof request. The operator fleet is externally operated
+/// and some endpoints have been observed accepting connections without ever
+/// responding; `reqwest::get` uses a client with no timeout, which would leave a
+/// validator hanging with no error at all. Mirrors `proposal_link.rs`.
```
MicaiahReid: This feels like a weird comment to be on a const. If massaged a bit, I think this would make more sense on `fn http_client`

svmgov/cli/src/utils/api_helpers.rs
```diff
@@ -1,11 +1,66 @@
...
+/// Reject a proof that describes a different account than the one requested.
+/// The leaf seeds the MetaMerkleProof PDA and is what the program verifies
+/// against the consensus root, so its identity is load-bearing — a mismatch
+/// would otherwise surface much later as an on-chain constraint failure.
+fn ensure_matches_request(
+    returned: &str,
+    requested: &str,
+    kind: &str,
+    base_url: &str,
+) -> Result<()> {
```
MicaiahReid: I think this comment explaining what the function does would make much more sense with either better var names (for `returned`/`requested`) and/or better types.

```rust
fn ensure_matches_request(
    requested_proof: &VoteAccountProofResponse,
    expected_vote_account: &str,
    proof_kind: &str,
    base_url: &str
)
```

Cause otherwise, the comment says it "rejects a proof that describes...", but it's really just comparing strings

svmgov/cli/src/instructions/cast_vote.rs
```diff
@@ -92,8 +92,11 @@ pub async fn cast_vote(
 
         let init_spinner = create_spinner("Initializing meta merkle proof...");
 
-        let voting_wallet = Pubkey::from_str(&proof_response.meta_merkle_leaf.voting_wallet)
-            .map_err(|e| anyhow!("Invalid voting wallet in proof: {}", e))?;
+        // Validate every field from the operator API rather than assuming it is
+        // well-formed: these converters return a clear error, where the const
+        // base58 decoder would panic on a malformed response.
```
MicaiahReid:
```suggestion
        // Validate every field from the operator API rather than assuming it is
        // well-formed
```
second half of comment seems irrelevant to new code

svmgov/cli/src/instructions/cast_vote_override.rs
```diff
@@ -118,8 +118,12 @@ pub async fn cast_vote_override(
     if meta_merkle_proof_account.is_none() {
         info!("Creating meta merkle proof account");
 
-        let voting_wallet = Pubkey::from_str(&meta_merkle_proof.meta_merkle_leaf.voting_wallet)
-            .map_err(|e| anyhow!("Invalid voting wallet in proof: {}", e))?;
+        // Validate every field from the operator API rather than assuming it is
+        // well-formed: these converters return a clear error, where the const
+        // base58 decoder would panic on a malformed response.
```
MicaiahReid:
```suggestion
        // Validate every field from the operator API rather than assuming it is
        // well-formed
```

MicaiahReid (changes requested):
Thanks @0xIchigo, just a few code style fixes!

MicaiahReid (approved):
Thanks @0xIchigo!

PR #153, opened 2026-08-14
fix(frontend): use proposal snapshot slot when generating proofs

### LINKED ISSUE #151: Frontend fetches proofs at the latest snapshot instead of each proposal's snapshot
*Note: this is an AI generated issue. First task: verify issue exists and severity*

---

## Summary

The frontend reads the verifier's latest `/meta` slot for vote and override proof requests, while `mapProposalDto` drops the selected proposal's `snapshotSlot`. That can make proof material come from a newer snapshot than the proposal's `consensusResult`, causing older active proposals to fail during vote submission.

### useSnapshotMeta.ts

Path: `frontend/src/hooks/useSnapshotMeta.ts`
Highlight lines: 2, 5

```typescript
// frontend/src/hooks/useSnapshotMeta.ts:10-24
queryFn: async (): Promise<NetworkMetaResponse> => {
  const url = `${ncnApiUrl}/meta?network=${network}`;
  ...
  return await response.json();
}
```

### useCastVote.ts

Path: `frontend/src/hooks/useCastVote.ts`
Highlight lines: 2, 4

```typescript
// frontend/src/hooks/useCastVote.ts:12-25
const { data: meta } = useSnapshotMeta();
...
castVoteMutation(params, { endpoint, network: endpointType, ncnApiUrl }, meta?.slot)
```

### getProposals.ts

Path: `frontend/src/data/getProposals.ts`
Highlight lines: 2, 4

```typescript
// frontend/src/data/getProposals.ts:80-86, 103-136
const consensusResult = rawAccount.account.consensusResult || undefined;
...
return { publicKey: rawAccount.publicKey, ..., consensusResult };
```

### castVote.ts

Path: `frontend/src/chain/instructions/castVote.ts`
Highlight lines: 5, 8

```typescript
// frontend/src/chain/instructions/castVote.ts:90-101
const voteAccountProof = await getVoteAccountProof(
  validatorVoteAccount.votePubkey,
  blockchainParams.network,
  slot,
  blockchainParams.ncnApiUrl
);
const metaMerkleProofPda = getMetaMerkleProofPda(
  voteAccountProof, SNAPSHOT_PROGRAM_ID, consensusResult
);
```

## Proof of Concept

The frontend uses the network-wide latest `/meta` slot for validator and delegator proof requests. Because the selected proposal's `snapshotSlot` is not retained, older active proposals can be assembled with proof material from a newer snapshot while still being bound to the proposal's `consensusResult`.

### useSnapshotMeta.ts

Path: `frontend/src/hooks/useSnapshotMeta.ts`

```typescript
// frontend/src/hooks/useSnapshotMeta.ts:10-24
queryFn: async (): Promise<NetworkMetaResponse> => {
  const url = `${ncnApiUrl}/meta?network=${network}`;
  ...
  return await response.json();
}
```

### useCastVote.ts

Path: `frontend/src/hooks/useCastVote.ts`

```typescript
// frontend/src/hooks/useCastVote.ts:12-25
const { data: meta } = useSnapshotMeta();
...
castVoteMutation(params, { endpoint, network: endpointType, ncnApiUrl }, meta?.slot)
```

### getProposals.ts

Path: `frontend/src/data/getProposals.ts`

```typescript
// frontend/src/data/getProposals.ts:80-86, 103-136
const consensusResult = rawAccount.account.consensusResult || undefined;
...
return { publicKey: rawAccount.publicKey, ..., consensusResult };
```

### castVote.ts

Path: `frontend/src/chain/instructions/castVote.ts`

```typescript
// frontend/src/chain/instructions/castVote.ts:90-101
const voteAccountProof = await getVoteAccountProof(
  validatorVoteAccount.votePubkey,
  blockchainParams.network,
  slot,
  blockchainParams.ncnApiUrl
);
const metaMerkleProofPda = getMetaMerkleProofPda(
  voteAccountProof, SNAPSHOT_PROGRAM_ID, consensusResult
);
```

## Recommendation

Preserve `raw.snapshotSlot` in `ProposalRecord` and pass that value to all proof and voter-summary requests. Use `/meta` only as a discovery/freshness endpoint. Before signing, optionally assert that both proof responses report the proposal slot and that the on-chain `ConsensusResult.snapshot_slot` matches it.

Fix checklist:

- [ ] Add `snapshotSlot` to `ProposalRecord` and populate it in `mapProposalDto`.
- [ ] Pass the selected proposal's `snapshotSlot` into every vote, override, and voter-summary proof request.
- [ ] Stop using the network-wide `/meta` slot as the proof slot source.
- [ ] Reject transaction assembly when a proof response's `snapshot_slot` does not match the selected proposal slot.

ChiefWoods:
Resolves #151 and #156

### DIFF
--- a/frontend/src/hooks/useCastVote.ts
+++ b/frontend/src/hooks/useCastVote.ts
@@ -2,25 +2,22 @@ import { CastVoteParams } from "@/chain";
 import { useEndpoint } from "@/contexts/EndpointContext";
 import { useNcnApi } from "@/contexts/NcnApiContext";
 import { castVoteMutation } from "@/data";
-import { resolveSnapshotNetwork } from "@/lib/snapshotNetwork";
+import { requireKnownSnapshotNetwork } from "@/lib/snapshotNetwork";
 import { useMutation } from "@tanstack/react-query";
 import { track } from "@vercel/analytics";
 
 export function useCastVote() {
-  const { endpointUrl: endpoint, endpointType } = useEndpoint();
+  const { endpointUrl: endpoint, network } = useEndpoint();
   const { ncnApiUrl } = useNcnApi();
 
   return useMutation({
     mutationKey: ["cast-vote"],
-    mutationFn: async (params: CastVoteParams) => {
-      const network = await resolveSnapshotNetwork(endpointType, endpoint);
-
-      return castVoteMutation(params, {
+    mutationFn: (params: CastVoteParams) =>
+      castVoteMutation(params, {
         endpoint,
-        network,
+        network: requireKnownSnapshotNetwork(network),
         ncnApiUrl,
-      });
-    },
+      }),
     onMutate: (params) => {
       track("Cast Vote init", { proposalId: params.proposalId });
     },
--- a/frontend/src/hooks/useCastVote.ts
+++ b/frontend/src/hooks/useCastVote.ts
@@ -3,33 +3,20 @@ import { useEndpoint } from "@/contexts/EndpointContext";
 import { useNcnApi } from "@/contexts/NcnApiContext";
 import { castVoteMutation } from "@/data";
 import { useMutation } from "@tanstack/react-query";
-import {
-  SNAPSHOT_UNAVAILABLE_MESSAGE,
-  useSnapshotMeta,
-} from "./useSnapshotMeta";
 import { track } from "@vercel/analytics";
 
 export function useCastVote() {
   const { endpointUrl: endpoint, endpointType } = useEndpoint();
   const { ncnApiUrl } = useNcnApi();
-  const { data: meta } = useSnapshotMeta();
 
   return useMutation({
     mutationKey: ["cast-vote"],
     mutationFn: (params: CastVoteParams) => {
-      if (meta?.slot === undefined) {
-        throw new Error(SNAPSHOT_UNAVAILABLE_MESSAGE);
-      }
-
-      return castVoteMutation(
-        params,
-        {
-          endpoint,
-          network: endpointType,
-          ncnApiUrl,
-        },
-        meta.slot
-      );
+      return castVoteMutation(params, {
+        endpoint,
+        network: endpointType,
+        ncnApiUrl,
+      });
     },
     onMutate: (params) => {
       track("Cast Vote init", { proposalId: params.proposalId });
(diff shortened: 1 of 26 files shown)


frontend/src/hooks/useCastVote.ts
```diff
@@ -3,33 +3,20 @@ import { useEndpoint } from "@/contexts/EndpointContext";
...
   const { endpointUrl: endpoint, endpointType } = useEndpoint();
   const { ncnApiUrl } = useNcnApi();
-  const { data: meta } = useSnapshotMeta();
 
   return useMutation({
     mutationKey: ["cast-vote"],
     mutationFn: (params: CastVoteParams) => {
-      if (meta?.slot === undefined) {
-        throw new Error(SNAPSHOT_UNAVAILABLE_MESSAGE);
-      }
```
MicaiahReid, line 22: So the code of this check doesn't actually make this clear, but `useSnapshotMeta` has `enabled: endpointType !== "custom"` in it.

This means, when a user is using a custom RPC network and we don't know if it's mainnet/testnet/devnet, we don't fetch snapshot meta. This is because the ncn service that is providing the snapshot needs the network type in the request.

So the `if (meta?.slot === undefined)` check here was doing more than it looks - it's checking if we can fetch a valid snapshot.

We can either keep this check as is (also for the other places that remove it such as `useCastVoteOverride`, etc), or have something like `if (endpointType === "custom")` instead.

---

On a side note, users should be able to go through the full flow with a custom RPC and still be routed through the ncn snapshot service appropriately. But that's a separate issue, tracked here: https://github.com/solana-foundation/solana-governance/issues/156

Even with this issue resolved, we will need to prevent voting for an unknown network type, so we'll need these checks back

MicaiahReid (changes requested):
Thanks @ChiefWoods! One change requested

frontend/src/hooks/useCastVote.ts
ChiefWoods: Makes sense, wouldn't have removed it if there's a comment clarifying it

ChiefWoods: Added checks for the affected hooks by comparing returned genesis hash with hardcoded values

MicaiahReid (changes requested):
This is still lacking if it's going to address #156 as well.

`useSnapshotMeta` is going to use `const { endpointType } = useEndpoint();`

If it's endpoint type is returned as "custom", it's going to throw an error. So even though `resolveSnapshotNetwork` works for the vote paths, `useSupportProposal` will still fail.

I think to properly address #156, we should likely have `EndpointContext` be upgraded to store type of mainnet/devnet/testnet/unknown. Custom URLs attempt to figure out network type, and default to unknown. If the `EndpointContext` is returning a network type that has been assumed based off the genesis hash, then `useSnapshotMeta` should work too.

MicaiahReid (approved):
Awesome work, thanks @ChiefWoods!

PR #113, opened 2026-08-03
fix(ncn): support Agave 4.2 snapshot generation

plsystems-dev:
## Summary

- pin the ledger stack to Agave 4.2.0
- port 4.2 account cleanup, storage ownership, shred-version, and background-service lifecycle changes
- select legacy or fixed-point stake math from the bank feature set
- propagate snapshot failures, preserve configured account paths, and copy only the required ledger range

## Validation

- `cargo +1.93.1 check --locked --workspace --all-targets`
- `cargo +1.93.1 test --locked -p cli -p verifier-service`
- `cargo +1.93.1 test --locked -p ncn-ledger forwards_account_paths_to_ledger_loading`
- isolated end-to-end run on a live Jito Agave 4.2.0-rc.0 validator: replayed slot 436924558, generated and reloaded the full snapshot with the same bank hash, and produced byte-identical MetaMerkle archives across two independent runs
- deployed the exact signed head on a live Jito Agave 4.2.0 validator; the validator process remained uninterrupted and the production verifier passed its health and operator checks
- isolated verifier upload using the deployed image: accepted an 89.3 MB signed archive, indexed 872 vote accounts and 1,319,532 stake accounts, and served vote, stake, and voter proofs
- verified the missing-slot path exits non-zero and the validator remained healthy throughout

PR #168, opened 2026-08-21
fix(ncn-ledger): pick the snapshot pair closest to the target slot

bola-sol:
### Problem

`await-snapshot` failed with `Blockstore missing data to replay to slot
440641000` on a node whose ledger fully covered the target.

The replay base is chosen as the highest full snapshot at or below the
target, then the highest incremental built on that full. `await-snapshot`
writes its own output snapshot into `--backup-snapshots-dir`, the same
directory it scans for bases, so a completed run leaves behind a full
snapshot with no incremental attached to it. The next run selects that
leftover and resumes from it instead of from the pair it just copied in.

#113 narrowed the `agave-ledger-tool blockstore copy` range to start at the
matched incremental, so the copied blockstore now covers only the slots
between that incremental and the target. A start slot pulled backwards by a
leftover snapshot falls outside the copied range and the pre-flight
`slot_range_connected` check fails.

### Fix

Score every full snapshot at or below the target by the slot its pair
actually reaches, and take the highest. An older full with a recent
incremental now beats a newer full with none.

`get_bank_from_ledger` and `load_and_process_ledger` had separate copies of
the two-step selection and had to stay in agreement, so the logic moved into
`snapshot_selection::select_snapshot_archives` and both call it. Mismatched
copies would mean the range check passes while the loader starts somewhere
else.

The chosen archives are now logged as `Selected full snapshot <slot> and
incremental snapshot <slot>`.

### Testing

`cargo test -p ncn-ledger` covers the leftover-snapshot case, halt-slot
filtering, a full snapshot with no incremental, and an empty result.

Verified manually against the directory that produced the failure: the
selector picks 440600281 plus 440640293 (707 slots) rather than the
leftover 440607616 (33,384 slots).


MicaiahReid:
@bola-sol there is a concern from Greptile, and a CI test won't run due to a lockfile issue - can you resolve?

bola-sol:
@MicaiahReid believe CI test should pass now. 

Race is a potential issue, but that's pre-existing code that this PR doesn't touch. The staging block looks like it came in `dee90ff` `git diff` on this shows no changes to it. This PR only replaces the archive selection above it.

PR #171, opened 2026-08-23
refactor(frontend): migrate web3.js based code to kit

ChiefWoods:
- replaces web3.js based handrolled instruction builders with kit-based Codama clients
- replaces old wallet adapter with Connectorkit
- fixes all unsafe BN to number type casts, excluding few safe bigint to number casts that are meant for bps operations
- uses `solana-program/stake` exported types

Preview of this PR: https://solana-governance-main-refactor.up.railway.app

Note: This PR is only one-half of a bigger frontend refactor, including only the Kit migration, without the UI improvements

Branch with UI improvements: https://github.com/ChiefWoods/solana-governance/tree/refactor


MicaiahReid (changes requested):
@ChiefWoods, if code is generated by codama, can we have it in a /generated folder? i.e. `ncn/generated/clients` or `ncn/generated`, etc

It's helpful for reviews now and in the future as the clients change.

frontend/src/components/proposals/detail/support-phase-progress/stats.ts
```diff
@@ -1,8 +1,10 @@
+import { formatBigintPercentage } from "@/helpers/bigint";
+
 export interface SupportStatsInput {
-  currentSupportLamports: number;
-  totalStakedLamports: number;
+  currentSupportLamports: bigint;
+  totalStakedLamports: bigint;
   /** Already resolved via `supportThresholdPercentFromConfig`. */
-  thresholdPercent: number;
+  thresholdPercent: string;
```
MicaiahReid, line 7: I'm curious why a string and not a number here

MicaiahReid: Other percents (like in `SupportStats` below) are numbers

frontend/src/components/proposals/detail/QuorumDonut.tsx
```diff
@@ -21,11 +22,11 @@ const defaultColors = {
 };
 
 interface QuorumDonutProps {
-  forLamports: number;
-  againstLamports: number;
-  abstainLamports: number;
+  forLamports: bigint | number;
```
MicaiahReid, line 25: should be these be the new `Lamports` type?

frontend/src/chain/quorum.ts
```diff
@@ -11,16 +11,18 @@ export const QUORUM_NUMERATOR = 1;
 export const QUORUM_DENOMINATOR = 3;
 export const QUORUM_FRACTION = QUORUM_NUMERATOR / QUORUM_DENOMINATOR;
 
+type Lamports = bigint | number;
```
MicaiahReid: Why are we keeping `number`, rather than just `bigint`?

frontend/src/components/TopVoters.tsx
```diff
@@ -106,8 +106,8 @@ export function TopVoters() {
     downloadCsvData(validators || []);
   };
 
-  const formatNumber = (num: number) => {
-    return num.toLocaleString();
+  const formatNumber = (num: bigint) => {
```
MicaiahReid: This should be called `formatBigInt`. Because a decimal bigint will lose some precision, can you also audit call sites to make sure this isn't a footgun?

`(10.1234).toLocaleString()` -> `'10.1234'`
`(101234n/10000n).toString()` -> `'10'`

MicaiahReid (changes requested):

frontend/src/components/proposals/detail/support-phase-progress/stats.ts
ChiefWoods: It's derived from a `cluster_support_pct_min_bps` which deserializes into bigint and kept as string. You could 'unsafely' cast it into a JS number but there's no guarantee that `cluster_support_pct_min_bps` would remain below 10000.

frontend/src/components/proposals/detail/QuorumDonut.tsx
ChiefWoods: Redundant, amended to only bigint

frontend/src/chain/quorum.ts
ChiefWoods: Redundant, amended to only bigint

svmgov/idl/codama.json
MicaiahReid, line 1: Can we get this file formatted? And ensure it's formatted when regenerated?

ncn/tests/generated.test.mjs
MicaiahReid, line 1: Let's remove this test in favor of: https://github.com/solana-foundation/solana-governance/issues/187

svmgov/tests/generated.test.mjs
MicaiahReid, line 1: Let's remove this test in favor of: https://github.com/solana-foundation/solana-governance/issues/187

frontend/src/components/Navbar.tsx
```diff
@@ -193,38 +187,13 @@ export default function Navbar() {
...
-                text="Connect Wallet"
-                onClick={() => setVisible(true)}
-              />
-            )}
+            <ConnectButton
+              className="rounded-full font-plus-jakarta-sans font-bold lg:h-9 lg:px-6"
+              showCluster={false}
+              showBalance={false}
+              showTokenList={false}
+              showTransactionHistory={false}
```
MicaiahReid: These are always false, which means there is a lot of boilerplate code that is never rendered. Can we slim down the connector kit code to get rid of this never-rendered stuff completely? `WalletDropDownContent.tsx` can get rid of props/code, `TokenListElement` can be deleted

frontend/src/components/connector-kit/common/WalletConnectQR.tsx
MicaiahReid, line 1: I don't see this QR being rendered - is this code reachable?

MicaiahReid (changes requested):

frontend/src/components/connector-kit/common/WalletConnectQR.tsx
ChiefWoods: Yes, would appear if a WalletConnect project ID is set in env. Also can be toggled in `ConnectorProvider.tsx`

ChiefWoods: Shall I remove this option?

MicaiahReid: I see - have you been able to test this flow? If not, I'd be in favor of removing for now and we can introduce later if requested

ChiefWoods: Yes, it's taken straight from [connector-kit](https://www.connectorkit.dev/), but removed for simplicity

frontend/src/components/connector-kit/common/SwapTokenIcon.tsx
MicaiahReid, line 1: Can be removed

frontend/src/components/connector-kit/common/utils.ts
```diff
@@ -0,0 +1,51 @@
+export const clusterColors: Record<string, string> = {
```
MicaiahReid: I don't see this referenced anywhere else in the code

frontend/src/components/connector-kit/common/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  type: string;
+  formattedTime: string;
+  instructionTypes?: string[];
+}) {
+  if (tx.type === 'program' && tx.instructionTypes?.length) {
+    const summary = tx.instructionTypes.slice(0, 2).join(' · ');
+    return `${tx.formattedTime} · ${summary}`;
+  }
+  return tx.formattedTime;
+}
```
MicaiahReid: I don't see these referenced anywhere else in the code

MicaiahReid: I don't see these referenced anywhere else in the code

frontend/next.config.ts
```diff
@@ -1,10 +1,20 @@
 import { withSentryConfig } from "@sentry/nextjs";
 import type { NextConfig } from "next";
-import path from "path";
 
 const nextConfig: NextConfig = {
-  /* config options here */
   cacheComponents: true,
+  turbopack: {
+    resolveAlias: {
+      "@solana/web3.js": "./src/lib/empty-solana-web3js.ts",
```
MicaiahReid: What is the purpose of this?

frontend/src/dummy-data/top-voters.ts
MicaiahReid, line 1: Doesn't look like any exports in this file are used

frontend/src/dummy-data/wallets.ts
MicaiahReid, line 1: This files exports are mostly not used. The only that is used is `WalletData`, which is imported by `governance.ts` to use in `GovernanceDashboardState`. However, `GovernanceDashboardState` isn't used anywhere - can you clean up those paths by removing while we're here? (This is legacy stuff - not yours - but since this PR is having to clean up types to match new format, might as well delete)

MicaiahReid (changes requested):

frontend/next.config.ts
ChiefWoods: Was a stub for connector-kit, removed

frontend/src/components/connector-kit/common/utils.ts
ChiefWoods: This is part of a UI change in another branch, will remove it

frontend/src/components/connector-kit/common/utils.ts
ChiefWoods: Also part of a UI change in another branch, removing it

PR #172, opened 2026-08-23
NCN verifier and router reliability test results

plsystems-dev:


Digitalenergyllc:
Adding the root cause behind the Digital Energy row in that table, since it turned out
to be a config trap rather than tuning — and one that any operator running the verifier
behind a reverse proxy can hit silently.

**Short version:** the 429s were a rate-limit *keying* failure, not limit values. My
verifier was already running at `GLOBAL_REFILL_INTERVAL=1` / `GLOBAL_RATE_BURST=60`
(6-10x more permissive than the 10/10 in `scripts/setup.sh`) and still 429'd, because
every client on the internet was sharing one bucket.

### What happens

`TrustedProxyKeyExtractor::extract()` only consults `CF-Connecting-IP` /
`X-Forwarded-For` / `X-Real-IP` when the connecting peer is inside
`TRUSTED_PROXY_CIDRS`; otherwise it falls back to the peer IP. That fallback is correct
and deliberate — the module docs are explicit about the forged-header threat.

My setup: nginx terminating TLS, container published as `-p 127.0.0.1:3000:3000` on
docker's default bridge, no Cloudflare (DNS-only record). I had set
`TRUSTED_PROXY_CIDRS=127.0.0.1/32`, which looks correct for a loopback-published
container.

It isn't. With docker port publishing, the container sees the connection sourced from
the **bridge gateway** (`172.17.0.1`), not `127.0.0.1`. So `is_trusted()` returned
false on every request, the forwarded-header branch never ran, and all traffic — router,
frontend, scanners — collapsed into a single bucket keyed on one address. nginx was
setting `X-Real-IP` and `X-Forwarded-For` correctly the whole time; the service just
never looked at them.

Fix was one line: ``` TRUSTED_PROXY_CIDRS=127.0.0.1/32,172.17.0.0/16 ```


### Evidence

A single-source load test can't distinguish a collapsed bucket from a working per-IP
one — both return 429 at volume. Two sources can: one host bursting, a second host
sending 1 request every 2 seconds, which no correctly-keyed per-IP limiter would ever
reject.

``` before: 429 429 429 429 429 429 429 429 429 429 429 429```

```after: 200 200 200 200 200 200 200 200 200 200 200 200 ```


Only change between runs was the CIDR line. Startup log confirms it:
`Loaded 1 trusted proxy CIDR(s)` → `Loaded 2`.

Re-measuring my row from the table on the same route, post-fix:

| Target | Requests | Successful proofs | Errors |
|---|---:|---:|---:|
| Digital Energy | 7 | 7 | 0 |

Verifier is live and serving round 1020 correctly post-change — `/meta` returns merkle root `2Jwjfi6KWQmqMraTkCcWqdj3VvyGcVgfNGYdzTQmaUj8` at slot 440641000, and `/proof/vote_account/...` serves normally.

### Two suggestions

**1. Make the misconfiguration visible.** A config that looks right and fails silently
is the worst case here. Either extend the default trusted set to include loopback +
RFC1918, or emit a warning when the observed peer address is outside the configured
trusted list — something like "peer 172.17.0.1 is not in TRUSTED_PROXY_CIDRS; rate
limiting by peer IP" on first occurrence. That single line would have made this a
30-second diagnosis.

Also worth noting `TRUSTED_PROXY_CIDRS` doesn't appear in the environment-variable table or the Rate Limiting section of the verifier-service docs. Since `make install-verifier-service` doesn't set it and the docs don't mention running behind a reverse proxy, an operator can follow the documented path exactly, add nginx for TLS, and end up here with no indication anything is off.


**2. `CF-Connecting-IP` precedence is a hazard for non-Cloudflare deployments.**
`client_ip_from_headers()` checks `CF-Connecting-IP` first, ahead of the rightmost
`X-Forwarded-For`. The XFF path is safe because the rightmost entry is appended by the
trusted proxy — but `CF-Connecting-IP` has no equivalent protection and outranks it. So
once an operator trusts their own reverse proxy, any client can forge `CF-Connecting-IP`
and mint a fresh bucket per request unless the proxy strips it. Operators genuinely
behind Cloudflare are fine, since CF overwrites it.

Worth documenting for the nginx path:

```nginx
proxy_set_header CF-Connecting-IP "";
```

Verified here: 200 requests each carrying a random forged `CF-Connecting-IP` produced an
identical 60/140 200-vs-429 split to the clean control, so the strip holds.

Happy to open a separate issue for either of these if that's easier to track.

PR #176, opened 2026-08-23
fix(frontend): preserve stake precision in merkle proofs

lostintime101:
## Summary

- parse proof `active_stake` fields as exact decimal strings before native `JSON.parse` can round them
- allow proof response stake values to remain lossless until Anchor encodes them as u64 values
- add a regression test using the live failing value above `Number.MAX_SAFE_INTEGER`

## Root cause

The verifier API serializes u64 lamport values as JSON numbers. The failing validator proof contains `active_stake: 16054077974334921`, which exceeds JavaScript's safe-integer range. Native `JSON.parse` silently changes it to `16054077974334920`; that one-lamport change mutates the MetaMerkleLeaf hash and causes `InitMetaMerkleProof` to fail with error 6009 (`InvalidMerkleProof`).

Reconstructing the reported live proof for proposal `AGHDQ6gjRFJPoyEcHuc4X7sbxJwyJfeKTb3UrGFzFNZD` confirms both console roots exactly:

- exact value `16054077974334921` -> `2Jwjfi6KWQmqMraTkCcWqdj3VvyGcVgfNGYdzTQmaUj8` (on-chain root)
- rounded value `16054077974334920` -> `C2y6YfyyQRoftanedZ4LcpnTvgwpmn9oKWzshsRt3Abc` (failed node)

The lossless handling is opt-in per response field, so slots and ordinary NCN response numbers retain their existing types.

## Testing

- `pnpm test -- --runInBand` (24 suites, 361 tests)
- `pnpm exec tsc --noEmit`
- ESLint on changed frontend files
- `pnpm build`

### DIFF
--- a/frontend/src/chain/instructions/helpers.ts
+++ b/frontend/src/chain/instructions/helpers.ts
@@ -170,6 +170,7 @@ export async function getVoteAccountProof(
 
   return fetchNcnJson<VoteAccountProofResponse>(url, {
     label: "vote account proof",
+    losslessIntegerFields: ["active_stake"],
   });
 }
 
@@ -184,6 +185,7 @@ export async function getStakeAccountProof(
 
   return fetchNcnJson<StakeAccountProofResponse>(url, {
     label: "stake account proof",
+    losslessIntegerFields: ["active_stake"],
   });
 }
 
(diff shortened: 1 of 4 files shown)


MicaiahReid (approved):
Good find, @lostintime101!

PR #181, opened 2026-08-23
fix(frontend): route plain vote signing through signTransactionForWallet

latent-9:
## What

The wallet-signing guard from #179 protects the override flows only. The plain cast vote, modify vote, support, and create proposal paths still send whatever `wallet.signTransaction` returns.

## Why

A wallet can return a transaction signed by a different account than the one that built it, for example when the connected account changes while the proof and blockhash fetches are in flight. The override flows detect this through `signTransactionForWallet` and raise a descriptive `WalletSigningError`. The plain validator flows sign directly on the wallet, so the same switch surfaces later as an opaque missing-signature serialization error after the user has already waited on every RPC round trip.

## Change

Capture the signer when the transaction is built and pass it to `signTransactionForWallet` in all four instruction builders, matching how `castVoteOverride.ts` and `modifyVoteOverride.ts` already work.

Tests for each builder mirror the override suite: an unsigned wallet response and a signature from a different account are rejected before submission, including a case where the account switches during one of the awaited fetches.

### DIFF
--- a/frontend/src/chain/instructions/__tests__/castVote.test.ts
+++ b/frontend/src/chain/instructions/__tests__/castVote.test.ts
@@ -1,0 +1,233 @@
+import { PublicKey, Transaction } from "@solana/web3.js";
+
+// helpers.ts transitively imports EndpointContext -> env.ts (an ESM-only package Jest does not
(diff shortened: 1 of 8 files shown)


MicaiahReid (approved):
Thanks @latent-9!

PR #182, opened 2026-08-24
fix(verifier): name SQL read columns

### LINKED ISSUE #161: Update verifier sql reads to name specific columns
all three reads use `SELECT *` with `row.get("column_name")`, which means that schema skew can turn into out-of-bounds panics instead of clean errors. It would be nice in a follow-up PR to name the columns explicitly, so it's a lot more robust for future migrations

_Originally posted by @0xIchigo in https://github.com/solana-foundation/solana-governance/issues/150#issuecomment-5332909334_

koriyoshi2041:
## Summary

- name every column read by the vote-account, stake-account, and snapshot metadata queries
- keep the returned records and filtering behavior unchanged

These reads previously used `SELECT *` and then accessed columns by name. Explicit projections make schema mismatches fail at query execution instead of letting stale row metadata reach infallible `row.get(...)` calls.

Fixes #161.

## Testing

- `rustfmt --edition 2021 --check ncn/verifier-service/src/database/operations.rs`
- `cargo test database` from `ncn/verifier-service` (12 passed)
- `git diff --check`

### DIFF
--- a/ncn/verifier-service/src/database/operations.rs
+++ b/ncn/verifier-service/src/database/operations.rs
@@ -92,7 +92,9 @@ impl VoteAccountRecord {
         snapshot_slot: u64,
     ) -> Result<Option<VoteAccountRecord>> {
         let row_opt = sqlx::query(
-            "SELECT * FROM vote_accounts \
+            "SELECT network, snapshot_slot, vote_account, voting_wallet, \
+                    stake_merkle_root, active_stake, meta_merkle_proof \
+             FROM vote_accounts \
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
Nice, thanks @koriyoshi2041!

MicaiahReid (changes requested):
@koriyoshi2041 Once again (as noted in [this surfpool PR](https://github.com/solana-foundation/surfpool/pull/732)), foundation PRs can't be merged with [commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) enabled. Please get this set up, then I can merge.

PR #184, opened 2026-08-24
fix(frontend): keep proposal countdown accurate

lostintime101:
## Summary
- estimate future epoch boundaries from 60 recent RPC performance samples, with a validated 350ms fallback
- refresh epoch data and the projected phase boundary every five minutes
- tick the visible countdown every minute without additional RPC calls
- avoid a server/client hydration mismatch by showing a placeholder until the client clock is available
- add unit coverage for weighted slot estimates, fallback behavior, and countdown ticks

## Context
The previous fixed 400ms estimate overstates phase duration now that mainnet slot production is faster. Recent observed samples were approximately 366ms per slot, enough to shift a multi-epoch proposal deadline by hours.

This is a more complete alternative to #183: that PR retains a 400ms fallback and does not refresh the projection or visible countdown after the initial query.

## Verification
- pnpm test: 29 suites, 387 tests passed
- pnpm exec tsc --noEmit
- ESLint on changed files
- pnpm build

### DIFF
--- a/frontend/src/helpers/date.ts
+++ b/frontend/src/helpers/date.ts
@@ -1,4 +1,46 @@
-import { Connection, EpochInfo, EpochSchedule } from "@solana/web3.js";
+import {
+  Connection,
+  EpochInfo,
+  EpochSchedule,
+  PerfSample,
+} from "@solana/web3.js";
+
+const DEFAULT_SLOT_TIME_MS = 350;
+const PERFORMANCE_SAMPLE_LIMIT = 60;
+const MIN_PLAUSIBLE_SLOT_TIME_MS = 100;
+const MAX_PLAUSIBLE_SLOT_TIME_MS = 1_000;
+
+/**
+ * Calculates a slot-duration estimate weighted by the number of slots in each RPC sample.
+ * Falls back to the configured 350ms slot time when an RPC omits or returns unusable samples.
+ */
+export function estimateSlotTimeMs(
+  samples: Pick<PerfSample, "numSlots" | "samplePeriodSecs">[],
+): number {
+  const totals = samples.reduce(
+    (acc, sample) => {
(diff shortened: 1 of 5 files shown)


frontend/src/helpers/date.ts
```diff
@@ -47,26 +89,38 @@ export async function epochToDate(
...
-    const blockTime = await connection.getBlockTime(epochInfo.absoluteSlot);
-    currentBlockTime = blockTime ? blockTime * 1000 : Date.now();
-  } catch {
+  // Anchor the projection to chain time and derive the current slot rate from the last hour of
+  // RPC performance samples. Slot production can run materially faster than the 400ms target,
+  // shortening epoch phases by hours; a fixed slot duration makes proposal countdowns stale.
+  const [blockTimeResult, performanceSamplesResult] = await Promise.allSettled([
+    connection.getBlockTime(epochInfo.absoluteSlot),
+    connection.getRecentPerformanceSamples(PERFORMANCE_SAMPLE_LIMIT),
+  ]);
```
MicaiahReid, line 100: This `epochToDate` fn is used in a few components throughout the frontend - should we be memoizing these requests to prevent these fetches from happening non-stop?

lostintime101: I think we already memoize them through TanStack Query

MicaiahReid: ah yes you're right, except now the `queryKey` in `useEpochToDate` has `absoluteSlot` in it:
```
queryKey: [
      "epochToDate",
      epoch,
      endpointUrl,
      epochData?.epochInfo.absoluteSlot,
    ],
```

Doesn't that mean the query will always re-execute, since slot is always increasing?

MicaiahReid (approved):

PR #209, opened 2026-08-30
fix(frontend): recover from concurrent proof initialization

### LINKED ISSUE #175: Frontend vote race: fetch shared proof if initialization fails
<a href="#">P1</a> **Concurrent proof initialization aborts**

If two delegators for the same validator and consensus result initialize the shared proof concurrently, the second initialization fails after the first creates the PDA, and this branch aborts instead of re-reading the now-existing proof, causing the second override transaction never to be submitted. The same race exists in `modifyVoteOverride`.

**Knowledge Base Used:**
- [NCN snapshot on-chain program](https://app.greptile.com/solana-foundation/-/custom-context/knowledge-base/solana-foundation/solana-governance/-/docs/ncn-snapshot-program.md)
- [Frontend chain integration](https://app.greptile.com/solana-foundation/-/custom-context/knowledge-base/solana-foundation/solana-governance/-/docs/frontend-chain-integration.md)

_Originally posted by @greptile-apps[bot] in https://github.com/solana-foundation/solana-governance/pull/174#discussion_r3838947785_

iamomm-hack:
Closes #175.

## Summary

Make meta-merkle-proof initialization idempotent for both override-vote flows. When two delegators share the same validator proof and submit concurrently, the transaction that loses the proof-PDA initialization race now re-reads confirmed chain state and continues if the winning transaction created the account.

## Problem

`castVoteOverride` and `modifyVoteOverride` use the same reusable meta-proof PDA. Before this change, both flows followed this sequence:

1. read the proof PDA and observe that it does not exist;
2. build and submit an `initMetaMerkleProof` transaction;
3. abort on any initialization confirmation error.

Two clients can both pass step 1 before either initialization lands. One transaction creates the deterministic PDA; the other fails because the account now exists. Treating that losing transaction as a fatal error prevents its otherwise-valid override vote from being submitted.

## Approach

- Add a shared `assertMetaMerkleProofInitialized` helper used by both cast and modify flows.
- Preserve the fast path after a successful initialization without an additional RPC read.
- On an initialization error, re-read the exact proof PDA at `confirmed` commitment.
- Continue only when the proof account is now present, which is the expected concurrent-winner outcome.
- Preserve the original initialization error when the account is still absent, so unrelated failures remain fail-closed.

This does not change proof derivation, proof contents, vote weights, transaction signing, or on-chain program behavior. It only makes the frontend's two-step initialization sequence safe under an idempotent race.

## Regression coverage

The tests cover all three helper outcomes:

- successful initialization performs no redundant account read;
- failed initialization plus a now-present proof is accepted;
- failed initialization plus an absent proof still throws.

The `castVoteOverride` integration-style unit test also models the complete race: the initial lookup returns `null`, initialization confirmation fails, the confirmed re-read finds the proof, and the vote transaction is still signed and submitted.

## Validation

- `pnpm run test:ci` — 29 suites passed, 396 tests passed
- `pnpm run lint` — passed
- `tsc --noEmit` — passed
- Focused override/helper suite — 2 suites passed, 23 tests passed
- `git diff --check` — passed

I also attempted `pnpm run build`; the local Windows install could not resolve multiple existing declared packages through its pnpm junctions and could not fetch the configured Google Fonts. No dependency or build-configuration files are changed in this PR, and the repository's frontend CI gate is the passing `pnpm run test:ci` command above.

### DIFF
--- a/frontend/src/chain/instructions/__tests__/castVoteOverride.test.ts
+++ b/frontend/src/chain/instructions/__tests__/castVoteOverride.test.ts
@@ -382,7 +382,7 @@ describe("castVoteOverride", () => {
     mockSendRawTransaction
       .mockResolvedValueOnce("losing-init-signature")
       .mockResolvedValueOnce("vote-signature");
-    mockConfirmTransaction.mockResolvedValue({
+    mockConfirmTransactionByPolling.mockResolvedValue({ 
       value: { err: { InstructionError: [0, "Custom"] } },
     });
 
--- a/frontend/src/chain/instructions/__tests__/castVoteOverride.test.ts
+++ b/frontend/src/chain/instructions/__tests__/castVoteOverride.test.ts
@@ -375,6 +375,25 @@ describe("castVoteOverride", () => {
     expect(mockSendRawTransaction).toHaveBeenCalledTimes(1);
   });
 
+  it("submits the vote when a competing transaction wins the proof initialization race", async () => {
+    mockGetAccountInfo
+      .mockResolvedValueOnce(null)
+      .mockResolvedValueOnce({ data: Buffer.alloc(0) });
+    mockSendRawTransaction
+      .mockResolvedValueOnce("losing-init-signature")
+      .mockResolvedValueOnce("vote-signature");
+    mockConfirmTransaction.mockResolvedValue({
+      value: { err: { InstructionError: [0, "Custom"] } },
+    });
+
+    await expect(
+      castVoteOverride(params, blockchainParams)
+    ).resolves.toEqual({ signature: "vote-signature", success: true });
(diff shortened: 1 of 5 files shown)


MicaiahReid (changes requested):
Thanks for the PR, @iamomm-hack! I'm about to review, but I won't be able to merge either way without commit signature verification enabled. Can you set that up and repush?

MicaiahReid:
Looks like we'll also need a rebase, @iamomm-hack

MicaiahReid (approved):

MicaiahReid (changes requested):
@iamomm-hack tests are now failing in CI, can you take a look?

MicaiahReid (approved):
Rad, thanks for the quick turnaround, @iamomm-hack!

PR #211, opened 2026-08-31
fix(frontend): Key `epochToDate` on the Epoch and Not the Slot

### LINKED ISSUE #189: Improve epochToDate memoization
> `queryKey` in `useEpochToDate` has `absoluteSlot` in it:
> ```
> queryKey: [
>       "epochToDate",
>       epoch,
>       endpointUrl,
>       epochData?.epochInfo.absoluteSlot,
>     ],
> ```
> 
> Doesn't that mean the query will always re-execute, since slot is always increasing? 

 _Originally posted by @MicaiahReid in [#184](https://github.com/solana-foundation/solana-governance/pull/184/changes#r3853724255)_

0xIchigo:
## TLDR

This PR closes #189. `absoluteSlot` in the query key meant every refresh created a new cache entry instead of reusing one, so `staleTime` never applied and `epochToDate` re-ran its RPC calls for an unchanged answer. This PR

## Problem

`useEpochInfo` has `refetchInterval: 5 * 60 * 1000`, so `absoluteSlot` changes every five minutes, and with it in the key every change is a new cache entry rather than a refetch:

- **`staleTime` was dead code.** A new key has no cached data, so it can never be served stale. The five-minute `staleTime` protected nothing
- **`epochToDate` is not cheap** — one or two RPC calls (`getBlockTime`, `getRecentPerformanceSamples`)
- A proposal detail page mounts **three** of these hooks (one in `phase-timeline`, two in `support-phase-progress`), so up to six RPC calls every five minutes
- Remounting always missed, since the slot had moved on
- Superseded entries accumulated until garbage collection

## Why this isn't just a revert

The key was `epochInfo.epoch` until #184 deliberately changed it:

```diff
-    queryKey: ["epochToDate", epoch, endpointUrl, epochData?.epochInfo.epoch],
+    queryKey: [..., epochData?.epochInfo.absoluteSlot],
```

Epochs last about two days, and each run re-reads the live slot rate — `date.ts` is explicit that "slot production can run materially faster than the 400ms target, shortening epoch phases by hours". Going back to the epoch alone would
recompute twice a week and reintroduce the stale countdown #184 fixed

The two issues pull in opposite directions, so the fix has to serve both

## Changes

```ts
queryKey: ["epochToDate", epoch, endpointUrl, epochData?.epochInfo.epoch],
staleTime: EPOCH_TO_DATE_REFRESH_MS,
refetchInterval: EPOCH_TO_DATE_REFRESH_MS,
```

`refetchInterval` recomputes in place rather than minting a new entry, so the refresh cadence #184 achieved as a side effect of the churn is preserved—now deliberate rather than accidental—while the cache stops churning and remounts serve from it

`epochInfo.epoch` stays in the key on purpose: once the target epoch is no longer in the future, `epochToDate` switches from projecting to reading the epoch's actual block time. That branch flip should not wait up to five minutes

## Tests

Three, in a new `useEpochToDate.test.tsx`:

- a slot advance alone does not recompute
- an epoch rollover recomputes immediately
- separate target epochs cache separately (the three-per-page case)

Regarding the other tests: `395 passed`, types clean, eslint clean, and prettier clean

### DIFF
--- a/frontend/src/hooks/useEpochToDate.ts
+++ b/frontend/src/hooks/useEpochToDate.ts
@@ -3,6 +3,9 @@ import { useEndpoint } from "@/contexts/EndpointContext";
 import { epochToDate } from "@/helpers/date";
 import { useEpochInfo } from "./useEpochInfo";
 
+/** How often the projection is redone to pick up a changed slot rate. */
+const EPOCH_TO_DATE_REFRESH_MS = 5 * 60 * 1000;
+
 /**
  * Hook to convert a Solana epoch number to a Date
  * @param epoch - The epoch number to convert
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useEpochToDate.ts
```diff
@@ -13,23 +16,26 @@ export function useEpochToDate(epoch: number | undefined) {
...
-      "epochToDate",
-      epoch,
-      endpointUrl,
-      epochData?.epochInfo.absoluteSlot,
-    ],
+    // The epoch, not `absoluteSlot`: a key that moves every refresh makes each
+    // one a new cache entry rather than a refetch, so `staleTime` never applies.
+    // The epoch still belongs here because `epochToDate` switches from
+    // projecting to reading actual block time once the target is no longer
+    // ahead, which should not wait for the interval below.
```
MicaiahReid: I don't like these sorts of comments in a repo - this is essentially Claude's justification on why to make the change and why _not_ the _previous_ code, not explaining the why of the code.

The code itself seems self-explanatory enough. Feel free to refactor the comment, but I'm in favor of removing.
```suggestion
```

MicaiahReid (changes requested):

MicaiahReid (approved):

PR #217, opened 2026-09-02
fix(frontend): Tolerate a Concurrently Created Meta Merkle Proof

### LINKED ISSUE #175: Frontend vote race: fetch shared proof if initialization fails

0xIchigo:
## TLDR

This PR closes #175. Every delegator under a validator shares one proof account, so creating it is a race. The override paths aborted instead of using the account that now exists, and the existence check treated a funded-but-uncreated address as
initialized

## The race

The PDA is `(MetaMerkleProof, consensusResult, meta_merkle_leaf.vote_account)`. A delegator derives it from the validator they delegated to, so a validator and all of their delegators contend on a single account. The issue describes two delegators: a validator voting while one of their delegators overrides, colliding identically

| Path | Creation | On collision |
|---|---|---|
| `castVoteOverride`, `modifyVoteOverride` | separate transaction | threw, override never submitted |
| `castVote`, `modifyVote` | bundled into the vote tx | whole vote fails atomically |

One correction to the issue: both self-heal on a manual retry, since the next attempt finds the account. So this is a cryptic failure needing a retry rather than a permanently blocked vote, though during a live vote plenty of delegators will read `Failed to initialize meta merkle proof` as "voting is broken"

Note that `skipPreflight` is never set, so preflight rejects the duplicate creation and `sendRawTransaction` throws rather than returning a signature. The existing `initConfirmation.value.err` guard is not on the path the race takes, so guarding it alone would not have fixed anything. The `try` spans sign, send, and confirm

Recovery is conditional on the account actually being present afterward, not on classifying an error, so a rejected signature or an underfunded payer still fails

## A second bug in the same check

`if (!merkleAccountInfo)` treated any account as initialized, including one holding only lamports. The address is derivable, so anyone can fund it, and `cast_vote_override.rs:116` requires the snapshot program to own it, so a dusted address skipped creation and failed the vote on chain. This is a clear griefing vector against the UI; the program was hardened against this in `prefunded_pdas_do_not_block_voting`, but the frontend was
not

`isMetaMerkleProofInitialized` checks owner and data, and all four vote paths use it. The existing tests encoded the old behaviour by mocking `{ data: Buffer.alloc(0) }`—exactly the dust shape—as initialized

## Reuse is safe

`init_meta_merkle_proof` verifies the leaf against the consensus root before returning, so an account created by another delegator cannot hold contents that would fail the vote

## Tests

- Seven new tests that cover race recovery, genuine failure still throws, dust triggers creation, and four direct cases on the predicate Verified by removal, not just by passing
- `420 passed`, types and eslint clean

### DIFF
--- a/frontend/src/chain/instructions/__tests__/castVote.test.ts
+++ b/frontend/src/chain/instructions/__tests__/castVote.test.ts
@@ -41,7 +41,7 @@ import { BN } from "@coral-xyz/anchor";
 import type { AnchorWallet } from "@solana/wallet-adapter-react";
 
 import { castVote } from "../castVote";
-import { SVMGOV_PROGRAM_ID } from "../types";
+import { SVMGOV_PROGRAM_ID, SNAPSHOT_PROGRAM_ID } from "../types";
 import type { RpcNetwork } from "@/types";
 
 // Distinct, valid 32-byte public keys used as stand-ins.
@@ -134,7 +134,12 @@ describe("castVote", () => {
       current: [{ nodePubkey: SIGNER, votePubkey: VOTE_ACCOUNT }],
       delinquent: [],
     });
-    mockGetAccountInfo.mockResolvedValue({ data: Buffer.alloc(0) });
+    // An initialized proof: owned by the snapshot program with data. Lamports
+    // alone would not qualify — see isMetaMerkleProofInitialized.
+    mockGetAccountInfo.mockResolvedValue({
+      owner: SNAPSHOT_PROGRAM_ID,
+      data: Buffer.alloc(32),
+    });
     mockGetLatestBlockhash.mockResolvedValue({
       blockhash: BLOCKHASH,
       lastValidBlockHeight: 123,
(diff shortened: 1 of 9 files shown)


0xIchigo:
whoops @MicaiahReid, just realizing that 209 also attempts to address 175. I'm good with closing my PR if needed

0xIchigo:
closing as 209 was merged

PR #226, opened 2026-09-14
fix(ncn/svmgov-program)!: convert BallotBox expiry to be slot-based

MicaiahReid:
This PR converts BallotBox expiry to be slot-based rather than reading `vote_duration` (in seconds) from a config. There was the possibility of a BallotBox being locked out.

Now, since all time-based math is in slots, slot-time changes in the future can all be addressed with config changes.

Note: this is a breaking change for the `BallotBox` account, but all existing accounts are already expired/reached consensus. As such, no migration path is needed


ncn/programs/ncn-snapshot/src/state/ballot_box.rs
```diff
@@ -27,9 +27,8 @@ pub struct BallotBox {
     /// one vote; tallies are not stake-weighted.
     #[max_len(MAX_BALLOT_TALLIES)]
     pub ballot_tallies: Vec<BallotTally>,
-    /// Timestamp when voting ends. Tie breaker admin will decide the results
-    /// if no consensus is reached by then.
-    pub vote_expiry_timestamp: i64,
+    /// Slot when voting ends.
+    pub vote_expiry_slot: u64,
```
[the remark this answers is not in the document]
MicaiahReid: There are no existing `BallotBox` accounts that aren't already expired.

ncn/programs/ncn-snapshot/src/lib.rs
```diff
@@ -59,8 +57,15 @@ pub mod ncn_snapshot {
         snapshot_slot: u64,
         proposal_seed: u64,
         spl_vote_account: Pubkey,
+        vote_expiry_slot: u64,
```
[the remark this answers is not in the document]
MicaiahReid: Fixed: [d08b65e](https://github.com/solana-foundation/solana-governance/pull/226/commits/d08b65e202fa9d825d0e3367e0b9d0c6b32f738a)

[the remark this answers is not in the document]
MicaiahReid: Can you verify this is fixed @greptileai

MicaiahReid:
@greptileai can you rereview

ncn/programs/ncn-snapshot/src/state/program_config.rs
```diff
@@ -20,8 +24,9 @@ pub struct ProgramConfig {
     pub min_consensus_threshold_bps: u16,
     /// Admin allowed to decide the winning ballot if vote expires before consensus.
     pub tie_breaker_admin: Pubkey,
-    /// Duration for which ballot box will be opened for voting.
-    pub vote_duration: i64,
+    /// Reserved to preserve the deployed account layout. This was the legacy
+    /// timestamp-based vote duration and is no longer read or configurable.
+    pub reserved: [u8; 8],
```
tiago18c, line 29: Should zero-out these bytes in the `update_program_config::handler` and make sure the instruction is run. 

Even though this has no benefit for now, it will make future upgrades easier if we ever need to repurpose those bytes as we can start with the assumption its zero rather than non-zero.

svmgov/program/programs/svmgov_program/src/utils.rs
```diff
@@ -165,15 +165,66 @@ pub fn is_valid_github_link(link: &str) -> bool {
...
+/// consensus before SVMGov voting opens.
+pub fn ensure_snapshot_before_voting_start(
+    snapshot_slot: u64,
+    voting_start_epoch: u64,
+) -> core::result::Result<u64, crate::error::GovernanceError> {
+    let voting_start_slot = epoch_start_slot(voting_start_epoch)?;
+    if snapshot_slot >= voting_start_slot {
+        return Err(crate::error::GovernanceError::SnapshotSlotNotBeforeVotingStart);
+    }
+    if voting_start_slot.saturating_sub(snapshot_slot) < ncn_snapshot::MIN_VOTE_EXPIRY_SLOTS {
```
tiago18c, line 222: this comparison makes the previous one redundant.

`snapshot_slot >= voting_start_slot` implies that the `saturating_sub` will be zero, so should we really keep both?

tiago18c (changes requested):
One nit and a possible simplification.

There are still some references left to  `vote_expiry_timestamp` in the docs (`ncn/QUICKSTART.md`, `docs/src/content/reference/errors/index.mdx` and `ncn/programs/ncn-snapshot/README.md`)

svmgov/program/programs/svmgov_program/src/utils.rs
MicaiahReid: Meh, I think it's nice to see this logic explicitly, and I think the different errors here are individually useful: snapshot slot bust be before voting starts, and must leave room for vote to happen

ncn/programs/ncn-snapshot/src/state/program_config.rs
MicaiahReid: Good call, done: 318429c5fc6b5eb968b531e703fa4ec8b0aad650

tiago18c (approved):

PR #236, opened 2026-09-16
feat(svmgov): support Ledger signers

plsystems-dev:
Add ledger support. 

tested via creating this Tx with ledger using this branch:

https://app.squads.so/squads/FQC6LwQSEPNEX7MrFJXSqty46r4wxA9iJdsqpzVsv8aD/transactions/BUWBj6TXFhVdkcsTgWuzukKJi8maeq9CBLY9mXxrpDU6


MicaiahReid (approved):
This looks great, thanks @plsystems-dev! Should we go ahead an update the NCN commands as well? I don't know if that generally helps NCN operators with their votes, but at least for the multisig/config updates it would be nice.

Happy to merge this and do a follow-up PR for NCN if you prefer

plsystems-dev:
> This looks great, thanks @plsystems-dev! Should we go ahead an update the NCN commands as well? I don't know if that generally helps NCN operators with their votes, but at least for the multisig/config updates it would be nice.
> 
> Happy to merge this and do a follow-up PR for NCN if you prefer

hey- i suggest we go ahead as is. Way less useful for NCN because we use a json validator identity (it cannot actually be ledger). but happy to support in a follow up either way.

PR #240, opened 2026-09-18
feat(ncn-router): set up IaC for router deployment

MicaiahReid:


infra/ncn-router/startup.sh.tftpl
```diff
@@ -0,0 +1,119 @@
...
+        proxy_read_timeout 65s;
+    }
+}
+NGINX
+
+ln -sfn /etc/nginx/sites-available/ncn-router /etc/nginx/sites-enabled/ncn-router
+rm -f /etc/nginx/sites-enabled/default
+
+systemctl enable --now docker.service
+systemctl disable --now nginx.service || true
```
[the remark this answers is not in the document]
MicaiahReid: Fixed: 634da68

.github/workflows/deploy-ncn-router.yaml
```diff
@@ -0,0 +1,208 @@
...
+              --tunnel-through-iap \
+              --ssh-key-expire-after=15m \
+              --quiet \
+              --command 'sudo /usr/local/sbin/ncn-router-deploy'
+
+      - name: Deployment summary
+        env:
+          IMAGE_REF: ${{ steps.image.outputs.ref }}
+        run: |
+          {
```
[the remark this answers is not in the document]
MicaiahReid: Fixed: 7ed8343

infra/ncn-router/scripts/ncn-router-deploy
```diff
@@ -0,0 +1,173 @@
...
+        /var/lib/ncn-router/ncn_whitelist.testnet.json >/dev/null 2>&1; then
+      break
+    fi
+    sleep 3
+  done
+
+  jq -e '.network == "mainnet" and (.verifiers | type == "array")' \
+    /var/lib/ncn-router/ncn_whitelist.mainnet.json >/dev/null
+  jq -e '.network == "testnet" and (.verifiers | type == "array")' \
+    /var/lib/ncn-router/ncn_whitelist.testnet.json >/dev/null
```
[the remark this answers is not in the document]
MicaiahReid: Fixed: 12176f7

infra/ncn-router/main.tf
```diff
@@ -0,0 +1,246 @@
...
+# Project scope supplies compute.projects.get for gcloud while OS Login and IAP
+# still determine whether the identity can enter a particular instance.
+resource "google_project_iam_member" "deployer_os_admin_login" {
+  project = var.project_id
+  role    = "roles/compute.osAdminLogin"
+  member  = local.deployer_member
+}
+
+# Instance scope prevents this otherwise broad predefined role from creating VMs.
+resource "google_compute_instance_iam_member" "deployer_instance_admin" {
```
tigarcia, line 220: **Blocking: these instance-scoped bindings are silently lost when the VM is replaced.**

Both `deployer_instance_admin` and `deployer_iap_ssh` key on `instance_name`/`zone`, which are identical before and after `replace_triggered_by` fires. When the VM is destroyed, its resource-level IAM policy goes with it; the new VM comes up with an empty policy, but Terraform sees no diff on these resources (same inputs) and doesn't recreate them. The binding exists in state and not in reality.

Right now this is masked by the "Verify idempotent second apply" step, which refreshes and recreates them — so it's invisible today, but it's a latent bug that would surface the moment that step changes (see my comment on the workflow).

Suggested fix, on both resources:

```hcl
lifecycle {
  replace_triggered_by = [google_compute_instance.router]
}
```

For the IAP tunnel binding I'm less sure IAP's resource model actually clears on VM deletion — could you confirm empirically after a forced replacement?

infra/ncn-router/main.tf
```diff
@@ -0,0 +1,246 @@
...
+  name         = local.name
+  project      = var.project_id
+  region       = var.region
+  address_type = "EXTERNAL"
+  network_tier = "PREMIUM"
+
+  depends_on = [google_project_service.required]
+}
+
+resource "google_compute_firewall" "cloudflare_https" {
```
tigarcia, line 61: **Blocking: Cloudflare-IP allowlisting isn't origin authentication.**

These ranges are shared by every Cloudflare customer, so anyone can create their own proxied hostname pointing at our origin IP and reach nginx directly — bypassing whatever WAF / rate-limit / bot rules we put on `ncn-governance.solana.com`. "Full (strict)" doesn't help here: it authenticates the origin *to Cloudflare*, not Cloudflare to us.

The standard fix is Authenticated Origin Pulls — in the nginx server block in `startup.sh.tftpl`:

```nginx
ssl_client_certificate /etc/ncn-router/tls/cloudflare-origin-pull-ca.pem;
ssl_verify_client on;
```

plus enabling AOP on the zone. Cloudflare's origin-pull CA is public, so it can be baked into the startup script rather than passed through Doppler.

Related: there's no `default_server` block, so any `Host` header lands on this vhost. A `return 444` default would tighten that too.

tigarcia (commented):
Really careful work overall — secrets never touch Terraform state or argv, digest-pinned images, SHA-pinned actions, strong container hardening, the port-22 IAP condition, a real rollback path, and CI asserting its own IAM blast radius. Nice.

I've left inline comments on the things I think should land before the first production deploy. Two I'd consider blocking (the IAM-binding-on-replacement bug, and origin authentication); the rest are should-fix. Happy to pair on any of them.

.github/workflows/deploy-ncn-router.yaml
```diff
@@ -0,0 +1,241 @@
...
+      - name: Validate Terraform
+        run: terraform -chdir="$TF_ROOT" validate
+
+      - name: Create saved plan
+        run: terraform -chdir="$TF_ROOT" plan -input=false -out="$RUNNER_TEMP/ncn-router.tfplan"
+
+      - name: Apply saved plan
+        run: terraform -chdir="$TF_ROOT" apply -input=false -auto-approve "$RUNNER_TEMP/ncn-router.tfplan"
+
+      - name: Verify idempotent second apply
```
tigarcia, line 93: **Should fix: this doesn't verify idempotency.**

`terraform apply -auto-approve` without a saved plan will plan, apply whatever it finds, and exit 0 — so if the first apply left drift, this step silently applies it and passes. To actually assert "nothing left to do":

```yaml
- name: Verify no drift after apply
  run: terraform -chdir="$TF_ROOT" plan -input=false -detailed-exitcode
```

(exit code 2 = changes pending → step fails).

Heads-up: as written this step is accidentally papering over the IAM-binding bug I flagged in `main.tf` — switching to `-detailed-exitcode` on its own would start failing after every VM replacement. The two changes need to land together.

infra/ncn-router/main.tf
```diff
@@ -0,0 +1,246 @@
...
+
+resource "google_artifact_registry_repository_iam_member" "runtime_reader" {
+  project    = var.project_id
+  location   = google_artifact_registry_repository.router.location
+  repository = google_artifact_registry_repository.router.name
+  role       = "roles/artifactregistry.reader"
+  member     = "serviceAccount:${google_service_account.runtime.email}"
+}
+
+resource "google_project_iam_member" "runtime_log_writer" {
```
tigarcia, line 126: **Should fix: `logWriter` is granted but nothing writes logs.**

Nothing in the startup script installs the Ops Agent or otherwise forwards journald / Docker logs to Cloud Logging, so this role is unused and the README's "writes Cloud Logging entries" isn't true yet. All router request logs and cron output stay on the VM — which means every VM replacement destroys the production log history, and we can't investigate an incident that preceded a deploy.

Either install `google-cloud-ops-agent` in `startup.sh.tftpl` (and point it at the docker/journald logs), or drop the role and adjust the README. I'd strongly prefer the former for a production endpoint.

infra/ncn-router/startup.sh.tftpl
```diff
@@ -0,0 +1,125 @@
...
+Wants=network-online.target ncn-meta-cron.service
+Requires=docker.service
+ConditionPathExists=/etc/ncn-router/deployment.env
+ConditionPathExists=/etc/ncn-router/runtime.env
+
+[Service]
+Type=simple
+EnvironmentFile=/etc/ncn-router/deployment.env
+ExecStartPre=-/usr/bin/docker rm --force ncn-router
+ExecStart=/usr/bin/docker run --rm --name ncn-router --user 10001:10001 --read-only --tmpfs /tmp:rw,noexec,nosuid,size=16m --cap-drop ALL --security-opt no-new-privileges --env-file /etc/ncn-router/runtime.env --publish 127.0.0.1:8080:8080 --volume /var/lib/ncn-router:/var/lib/ncn-router:ro $${NCN_ROUTER_IMAGE} /usr/local/bin/ncn-router
```
tigarcia, line 73: **Should fix: two independent disk-fill paths on a 20 GB disk.**

1. Docker's default `json-file` log driver has no size cap, and `ncn-router` logs one line per request to stderr.
2. `ncn-router-deploy` pulls a new digest every deploy and never prunes, so every image ever deployed accumulates.

Suggest writing `/etc/docker/daemon.json` in the startup script before enabling docker:

```json
{ "log-driver": "journald" }
```

(or `json-file` with `"log-opts": {"max-size": "50m", "max-file": "5"}`), and adding a `docker image prune --force --filter until=…` or equivalent to the deploy script after readiness passes. `journald` also gets us most of the way to shipping logs to Cloud Logging via the Ops Agent.

.github/workflows/deploy-ncn-router.yaml
```diff
@@ -0,0 +1,241 @@
...
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Load deployment credentials and runtime secrets from Doppler
+        uses: dopplerhq/secrets-fetch-action@451892f16195f9ac360e1a5bcbf0b5fd0e957534 # v2.0.0
+        with:
+          auth-method: oidc
+          doppler-identity-id: ${{ vars.DOPPLER_SERVICE_IDENTITY_ID }}
+          doppler-project: solana-governance
+          doppler-config: prd
+          inject-env-vars: true
```
tigarcia, line 145: **Should fix: the deploy job inherits every secret in Doppler `prd`.**

`inject-env-vars: true` exposes the whole config to this job, and `prd` is shared with `release-programs.yaml` (Squads vault / multisig material, etc.). You already did the right thing by carving out `prd_infra` for the Terraform job — the same argument applies here. A dedicated config (e.g. `prd_ncn_router`) holding just the deploy key, RPC URLs, and Origin CA material would mean a compromise of this job can't reach governance-release secrets.

infra/ncn-router/main.tf
```diff
@@ -0,0 +1,246 @@
...
+  project = var.project_id
+  role    = "roles/logging.logWriter"
+  member  = "serviceAccount:${google_service_account.runtime.email}"
+}
+
+# Bootstrap changes must create a new VM. Google's Compute Engine will
+# update startup metadata in place, but only executes the change during 
+# boot. The static IP is a separate resource and remains attached to 
+#the replacement instance.
+resource "terraform_data" "router_bootstrap" {
```
tigarcia, line 136: **Should fix / discuss: VM replacement is an outage with no rollback target.**

Any edit to `startup.sh.tftpl` *or* `scripts/ncn-router-deploy` (base64'd into the startup script) recreates the VM. The new instance boots with no `deployment.env`, so the startup script deliberately leaves nginx disabled — the site is down from the moment Terraform replaces the instance until the `deploy` job's SSH step succeeds. If that job fails, the "previous deployment" backup that the rollback logic relies on was on the VM we just deleted.

I don't think that necessarily needs solving in this PR, but I'd like it to be a documented, deliberate trade-off (README + a note that startup-script edits are outage-inducing changes), and worth a follow-up issue for either a MIG or a create-before-destroy flow with the static IP swapped over after readiness.

infra/ncn-router/main.tf
MicaiahReid: Fixed: aa8d1a525d90b32f7b6a5898c90ae925aa1b0bfd

infra/ncn-router/main.tf
MicaiahReid: Fixed: ff1132bc79999f034bd8f0aa9938a55059df2e23

infra/ncn-router/main.tf
MicaiahReid: Fixed: cd2a28ada0b29ed4b5267b2ea824a917b4a2a869

infra/ncn-router/startup.sh.tftpl
MicaiahReid: Fixed: 1a3f4cad7ddaee0461a0d3304a82cc2b0ce6e51d

.github/workflows/deploy-ncn-router.yaml
MicaiahReid: Fixed: aa8d1a525d90b32f7b6a5898c90ae925aa1b0bfd

